PR #214, opened 2026-09-10
Derive genesis module registration from one table

paxlabs-inc:
## Issue

Genesis module registration was restated independently in six places. `src/protocol/lxp_genesis.c` (`lxp_genesis_state_root`, `lxp_genesis_materialize`), `cmd/layerx-genesis/lxp_genesis_builder.c`, `cmd/layerxd/lxp_daemon_process.c`, `cmd/layerx-guarantor/runtime.c` and `cmd/layerx-module-registry/main.c` each carried their own ordered sequence of `lxp_kernel_register_module` calls with their own version and custody conditions, the module count was a literal repeated alongside them, and Escrow, Budget, Stream, Service and Perps had no way to reach genesis at all. The registry sequence had already drifted: it omitted Governance from the document that `platform/hosted/tests/beta-cluster.sh` `module_registry_verify` compares against the daemon's LNI-advertised module list.

## What changed and why

**One table.** `src/protocol/lxp_genesis.c:39` holds `module_table[]`, one row per module id: the id, the gate that decides when the row applies, a default-enabled flag and the iface constructor.

| module | gate | default |
|---|---|---|
| Programs | `ALWAYS` | on |
| Asset | `STATE_COMMITMENT` | on |
| Governance | `STATE_COMMITMENT` | on |
| Bridge | `CUSTODY_PROFILE` | on |
| Escrow, Budget, Stream, Service, Perps | `ENABLE_FLAG` | off |

**One resolver.** `lxp_genesis_module_plan_resolve` turns a manifest into an `lxp_genesis_module_plan`; `lxp_genesis_module_plan_register` registers it; `lxp_genesis_module_plan_matches` checks a kernel against it. Every consumer now calls those instead of open-coding a sequence, so the module count and the registration order are derived from the table rather than repeated. `cmd/layerx-module-registry` derives its document from `lxp_genesis_module_plan_default`, which is why it now advertises Governance and matches the daemon. The daemon hunk is confined to the genesis registration section: `load_genesis_settlement_anchor` fills a plan next to the custody profile it already reads, and `open_process` registers it.

**Enable flags.** An `ENABLE_FLAG` row reads a genesis governance parameter whose key is the 32-byte zero-padded string `module-enable:` followed by the module iface name, and whose value is thirty-one zero bytes then `0` or `1`. `lxp_genesis_module_enable_key` is the only constructor. A duplicate parameter is refused as `LXP_ERR_SEQUENCE_REUSED`, any other value shape as `LXP_ERR_NON_CANONICAL`, and a `module-enable:` parameter naming a module without an `ENABLE_FLAG` row as `LXP_ERR_UNKNOWN_MODULE` rather than ignored. A duplicate module id in the table is refused by `module_table_validate`.

**Defaults preserve the current chain.** The five new rows default to disabled, so the resolved plan, the projected genesis state root, the materialization checks and every existing snapshot are byte-identical to the previous hardcoded Programs + Asset + Governance + optional Bridge sequence. No genesis fixture needed regeneration. Turning a module on is a one-line change of the `default_enabled` field in the row that already exists.

**Tests.** `tests/test_genesis_module_table.c` checks the table shape and per-row iface agreement, the four gates against every supported protocol version and the custody profile, the enable-key construction and each of its refusals, and then builds real signed manifests through `lxp_genesis_build_fresh_empty` to verify that enabling a flag-gated module changes the genesis state root, that the plan registers into a real kernel with the incremental `lxp_snapshot_load` refusals intact, and that the manifest signature still verifies. `tests/test_genesis_manifest.c` now drives its registrations through the plan and keeps its existing refusals.

## Gates

|---|---|---|---|---|
| build | `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/build.log` | `0384d5a35` |
| ledger | `make beta-ledger-check` | 0 | `qual-logs/ledger-check.log` | `0384d5a35` |
| test | `make CC=gcc test` | 0 | `qual-logs/test.log` | `0384d5a35` |
| genesis | `make CC=gcc test-genesis-bootstrap` | 2 | `qual-logs/test-genesis-bootstrap.log` | `0384d5a35` |
| registry | `make CC=gcc test-module-registry` | 2 | `qual-logs/test-module-registry.log` | `0384d5a35` |

`make CC=gcc test` does not reach `test-genesis-bootstrap` or `test-module-registry`; both were run explicitly because they cover the changed paths. `test-genesis-bootstrap` stops at its first binary, so each of its four binaries was also run directly at the same revision: `build/tests/test_genesis_module_table` exits 0 (`qual-logs/test_genesis_module_table.log`) and `build/tests/test_daemon_bootstrap_artifact` exits 0 (`qual-logs/test_daemon_bootstrap_artifact.log`), while `build/tests/test_genesis_manifest` and `build/tests/test_genesis_builder_cli` exit 1 (`qual-logs/test_genesis_manifest.log`, `qual-logs/test_genesis_builder_cli.log`). Those two, and `test-module-registry`, fail on three checks that were already failing on the base revision. Each was reproduced on a pristine tree with every change of this branch stashed, and recorded rather than repaired. Comparison logs are `qual-logs/test-genesis-manifest-main.log`, `qual-logs/test-genesis-builder-cli-main.log` and `qual-logs/module-registry-main.log`.

## Observations added

- `observation.6.11.2600` — the table, the gates, the enable-flag encoding and its refusals, and the decision to default the five new rows off so the genesis root is unchanged.
- `observation.6.11.2601` — the registry previously omitted Governance while the daemon registered and advertised it; deriving the registry from the shared plan removes the divergence.
- `observation.6.11.2602` — pre-existing: `cmd/layerx-module-registry/test_registry.py:24` expects Asset ordinals 1 through 9 while `src/modules/asset/lx_asset_registry.c:11-16` advertises eleven.
- `observation.6.11.2603` — pre-existing: `tests/test_genesis_manifest.c:203` requires Asset ordinal 9 not to dispatch while the registry advertises `LX_ASSET_WITHDRAW` as ordinal 9.
- `observation.6.11.2604` — pre-existing: `lxp_genesis_verify_signature` on `tests/fixtures/pay1-retired-genesis.manifest` returns `LXP_ERR_ROOT_MISMATCH`, so `test_genesis_builder_cli` exits 1 before it reaches the migration path.


paxlabs-inc:
Rebased d56f17c44 onto main c567bd265 after 0384d5a35: the only difference from main is the appended ledger blocks and the eleven files listed in the body; no C source on main changed between the two bases, so the build and test gates in the body hold for this head. Rerun at d56f17c44: make beta-ledger-check exit 0, git diff --check exit 0, and git diff origin/main HEAD -- spec/layerx-beta/qualification.kvx shows zero deleted lines.

PR #216, opened 2026-09-10
Hold the treasury key in a signer process behind a unix socket

paxlabs-inc:
## Issue

The node bootstrap wrote the treasury Ed25519 seed to `$DATA_DIR/secrets/treasury-key.hex` and named that path in a world-readable `node.env`, so every container sharing the node data directory held the treasury identity. This change puts the treasury key behind a signer process that speaks over a unix socket, stops the bootstrap from copying the seed anywhere, makes `node.env` owner-only, and leaves a provider interface so a key management service can hold the identity later.

## What changed

**`platform/hosted/node/signer/` (new)**

* `signer.py` — the treasury signer. It listens on a unix socket (mode 0600, or 0660 with `--socket-group`), admits peers by `SO_PEERCRED` against `--allowed-uid`, and answers two requests: `public-key` returns the treasury public key and its DID, `sign <64 hex>` returns an Ed25519 signature over that 32-byte digest. It pins the public key at start, re-checks it under a lock before every signature, and verifies each signature against that key before returning it, so a provider that rotates or forges is refused rather than trusted. Requests are bounded, malformed ones are refused with an explicit code, and the socket is unlinked on `SIGTERM`/`SIGINT`. A peer refused by its credentials is answered before it has finished writing, so the signer reads the bytes still in flight off the connection instead of closing over them and resetting away the refusal.
* `provider.py` — the provider interface behind the signer. The built-in `file` provider opens material with `O_NOFOLLOW`, requires a regular single-link file owned by the effective uid or root that is neither readable by others nor writable by its group, accepts 32 raw bytes or 64 hex characters, and loads it into an anonymous memory file so the seed is never copied onto a filesystem. The `command` provider delegates `public-key` and `sign <hex>` to an external program with a bounded output and a deadline, which is where a key management service plugs in.
* `client.py` — the client the rest of the node uses, and a small CLI for the same two requests.

**`platform/hosted/node/bootstrap.sh` (treasury hunks)**

`--treasury-signer-socket PATH` is now the exclusive alternative to `--treasury-key`; one of the two is required. With a socket, the bootstrap asks the signer for the public key rather than deriving it from a seed. It no longer writes `secrets/treasury-key.hex`, no longer emits `LAYERX_NODE_TREASURY_KEY_FILE`, emits `LAYERX_NODE_TREASURY_SIGNER_SOCKET` instead, publishes `LAYERX_CORE_TREASURY_SIGNER_SOCKET` into `core.env`, and writes `node.env` mode 0600 instead of 0644.

**`platform/hosted/node/supervisor.sh` (treasury hunk)**

Waits for the treasury signer socket before it bootstraps, bounded by `LAYERX_NODE_TREASURY_SIGNER_WAIT_SECONDS` (default 600), and carries the socket path into `core.env` when it republishes the core environment.

**`platform/hosted/node/deployment.yaml` (signer container block)**

A `treasury-signer` container runs next to `layerxd` as uid 4020 with a read-only root filesystem and all capabilities dropped. The `layerx-node-keys` secret is now projected twice: the `keys` volume selects `sequencer.key` alone for `layerxd`, and a `treasury-material` volume selects `treasury.key` alone for the signer and for the core boundary, which still reads it from a file. The daemon container therefore no longer mounts the treasury seed. `layerxd` takes `--treasury-signer-socket` in place of `--treasury-key`, and the node image carries the signer package at `/opt/layerx/signer`.

**Consumers**

The cluster deployment producer in `platform/hosted/tests/beta-cluster.sh` reads an owner-only `node.env`, asserts it names no treasury key material, and asks the signer for the public key and for the signature over the activity preimage. `platform/hosted/registry/tests/native_deployment.rs` starts a real signer over the cluster treasury material and points the producer environment at its socket.

**Ledger and report**

Ten observations are appended to `spec/layerx-beta/qualification.kvx`, and `spec/layerx-beta/report.md` is re-rendered from it by `tools/ci/beta-report.sh` in the same commit: the ledger digest, the observation totals and the task 6.1 blocker count move by exactly these ten records.

**Tests**

`platform/hosted/node/tests/signer-test.py` covers the socket contract and the DID, raw and hex material, refusal of material that is readable by others, group-writable, short, a symlink, a directory or absent, exclusive provider selection, bounded and malformed requests, the command provider including a forged signature and a rotated key, peer-credential admission and refusal across a second uid, concurrent signing, public key publication, and the bootstrap treasury boundary end to end against a real signer, the native binaries and the repository settlement document unmodified (its `certificate_threshold` of 2 now bootstraps, so the test no longer rewrites it to one and asserts instead that the shipped threshold is above one): the socket path reaches `node.env` and `core.env`, `LAYERX_NODE_TREASURY_KEY_FILE` is gone, `node.env` is 0600, the treasury identity is registered, and no file under the data directory contains the seed in either form.
## Gates

Head `202543a27e7d3699ed35bc2fe0572303c2c23063` on `origin/main` `dfd9618fe`.

Re-run at this head after the rebase (`tests/support/lxgb_metadata.py` arrived on main with PR 248, which closes the two cases that were red):

| Command | Exit | Log | Revision |
|---|---|---|---|
| `python3 platform/hosted/node/tests/signer-test.py -v` — **14 of 14** | 0 | `qual-logs/signer-test.log` | `dd1f59a28` |
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/native-build.log` | `dd1f59a28` |
| `make beta-ledger-check` | 0 | `qual-logs/ledger-check.log` | `202543a27` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `202543a27` |
| `git diff --check origin/main HEAD` | 0 | `qual-logs/diff-check.log` | `202543a27` |

`dd1f59a28` is this head before two amendments that touched only `spec/layerx-beta/qualification.kvx` (the wording of observation `6.1.4809`) and the `spec/layerx-beta/report.md` it renders; neither file is an input to the signer suite or the native build.

Run at `c45638836`, the previous head on `origin/main` `e28541ee8`, and not re-run after this rebase:

| Command | Exit | Log |
|---|---|---|
| `make CC=gcc test` | 0 | `qual-logs/native-test.log` |
| `python3 platform/hosted/tests/topology_check.py` | 0 | `qual-logs/topology-check.log` |
| `bash -n` on bootstrap.sh, supervisor.sh, beta-cluster.sh; `python3 -m py_compile` on the signer package and its test; python yaml load of deployment.yaml | 0 | `qual-logs/static-check.log` |
| `cargo test --offline --manifest-path platform/Cargo.toml --locked -p layerx-platform-registry --test native_deployment --no-run` | 0 | `qual-logs/registry-compile.log` |
| `cargo test ... --test native_deployment -- --test-threads 1` | 101 | `qual-logs/registry-producer.log` |
| `make CC=gcc platform-test-node` | 2 | `qual-logs/platform-test-node.log` |

The two red ones are red before any code this change touches.

**`platform-test-node`** — passes its `bash -n` step and its `layerxctl` build and tests, then stops in `cargo --locked` on a stale probe lock before any node starts. `platform/hosted/node/tests/probe/Cargo.lock:644` records `layerx-client` without `sha2`, which `agent/crates/layerx-client/Cargo.toml:13` requires; the probe is a private workspace outside the platform and agent workspaces, so no ordinary gate relocks it. `platform/hosted/node/tests/probe/`, `platform/hosted/node/tests/node-test.sh` and `platform/Makefile.inc` are byte-identical to `origin/main` on this branch (`6.1.4803`, `6.1.4808`).

**`native_deployment`** — fails in `start_cluster` with `layerxd: bootstrap configuration failed with result -402`, on all three cases including the two that contain no treasury or signer code path. `build_genesis` in that harness builds its own request and does not read the shared metadata fixture, so it is a distinct cause from the signer suite (`6.1.4804`).

## Observations added

* `6.1.4800` (note) — no file under `cmd/layerxd` names the treasury; the component that signs with it is the hosted core at `platform/hosted/core/src/main.rs:259`. The bootstrap publishes `LAYERX_CORE_TREASURY_SIGNER_SOCKET` for it and the core boundary keeps its file projection until that crate gains a client.
* `6.1.4801` (blocker, superseded by `6.1.4809`) — the genesis request length assertion expected `395 + metadata` bytes while the request carries one 81-byte guarantor entry per guarantor, so the repository settlement document's `certificate_threshold` of 2 refused.
* `6.1.4802` (blocker) — `node-test.sh`, `bootstrap-test.py` and the manifest omit `--genesis-metadata`, which the bootstrap requires.
* `6.1.4803` (blocker) — `platform/hosted/node/tests/probe/Cargo.lock` is stale by nineteen packages and `--locked` refuses it, stopping `platform-test-node` at its first step.
* `6.1.4804` (blocker) — `start_cluster` in `native_deployment.rs` cannot start layerxd; all three cases fail identically, including the two that touch no signer path.
* `6.1.4805` (note) — the Dockerfile line that carries the signer package into the image, and the split treasury projection that leaves the core boundary reading a file.
* `6.1.4806` (note) — `spec/layerx-beta/report.md` records a sha256 over the ledger plus per-severity and per-task counts, so two branches that each append observations conflict on those lines and no textual resolution is correct; the merge is resolved by re-running `tools/ci/beta-report.sh` on the merged ledger.
* `6.1.4807` (blocker, superseded by `6.1.4809`) — the shared genesis metadata fixture emitted an asset record `issuer_valid` refuses, so `layerx-genesis-build` refused every request built from it.
* `6.1.4808` (blocker) — the specific cause behind `6.1.4803`: `layerx-client` gained `sha2` and the probe's private workspace lock still records it without one, so `--locked` refuses the probe build before any node behaviour runs.
* `6.1.4809` (note) — both refusals are closed at this revision and the suite runs fourteen of fourteen. The length assertion now reads `314 + 81 * GUARANTOR_COUNT + metadata`, so the shipped settlement document bootstraps and the test passes it unmodified; the shared fixture now emits the accepted asset record shape, so both end-to-end bootstrap cases build genesis, register the treasury identity and prove no copy of the seed survives. Neither bound nor fixture was changed from this branch; both are byte-identical to `origin/main` here. This supersedes `6.1.4801` and `6.1.4807`.


paxlabs-inc:
Rebased mechanically onto main e33074328: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head fcbd2a510: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 202543a27 (no file overlap with main since the merge base).

PR #221, opened 2026-09-10
Dispatch every budget ordinal against canonical module state

paxlabs-inc:
## Issue

The budget module advertised seven activity types through
`lx_budget_module_iface()` but had no dispatch behind them: no `decode`,
no `validate`, no `execute`, and no canonical module state. Its
value-moving helpers were reachable only from tests, and every transfer
set they built left `context.source_authorities` NULL with
`source_authority_count` zero, which `lxp_kernel_apply_transfer_set`
refuses outright with `LXP_ERR_NON_CANONICAL`. Four of the five budget
targets failed on an unmodified checkout: `test_budget_create`,
`test_budget_spend`, `test_budget_delegate` and `test_budget_close` each
exited 1; only `test_budget_period` passed.

## What changed and why

**Real dispatch for every advertised ordinal.**
`src/modules/budget/lx_budget_module.c` implements the module interface
for all seven: create, fund, amend, delegate add, delegate remove, spend
and close. `decode` arena-allocates the typed payload and refuses an
ordinal outside 1..7; `validate` requires the activity type to equal
`(LXP_MODULE_BUDGET << 16) | ordinal` and charges gas; `execute`
authorises the signer and then runs the ordinal.

**Canonical key-value state.** Budget records now live in module state
under the 39-byte key `"budget:" || budget_id`.
`src/modules/budget/lx_budget_codec.c` encodes a versioned fixed layout
with strictly ascending delegates, validates on both encode and decode,
and carries the six payload codecs. The runtime store remains a read
fallback so the existing store-path helpers keep working.

**The transfer refusal is fixed at its cause.**
`lx_budget_bind_source_authority` builds the one-entry source authority
table the kernel requires and is called from every emission site in the
module. Funding uses `LXP_AUTH_OWNER` with the owner as the debited
account; spending and draining use `LXP_AUTH_BUDGET_ALLOWANCE` with the
budget account, which is what `lx_budget_authority_check` demands of an
`LX_ACCOUNT_AGENT_BUDGET` source in `src/ledger/lxp_apply.c`. No
assertion in the kernel or the journal was relaxed.

**The signer is bound to the activity.** `budget_authorized_signer`
requires owner authority, `authority.verified_key` to equal the activity
authority bytes, the derived DID id to equal `authority.actor`, a
verified payload hash and signature, and then requires the resolved
principal account to be the actor's own `agent:<did>:main` with a
matching `authority_key`, and `account_sequence` to equal that account's
`next_sequence`. Create additionally requires the payload's budget
account to be an `LX_ACCOUNT_AGENT_BUDGET` account under the same actor,
so a budget record can no longer be planted over another agent's budget
account.

**Events.** Each ordinal emits one event with a fixed body: create
(type 1, 80 bytes), fund (2, 48), amend (3, 56), delegate add (4, 64),
delegate remove (5, 64), spend (6, 80), close (7, 56).

**Refusals.** Unknown budget, closed budget, wrong owner, unregistered
delegate, allowance exceeded, insufficient budget funds, expired,
revoked, stale revocation, period cap on amend, non-canonical account
kinds, and the envelope refusals (bad signature, payload hash mismatch,
truncated payload, zero amount, sequence reuse, sequence gap).

**Tests.** `tests/modules/test_budget_create.c` keeps its store-path
coverage unchanged and adds a dispatch suite that drives all seven
ordinals through `lxp_kernel_dispatch` with ed25519-signed activities
over real accounts opened in an `lx_account_registry`. It asserts
balances, event bodies, the record read back through
`lxp_state_proof_build` / `lxp_state_proof_verify` against the live state
root, capped rollover across a period boundary, delegate admission and
removal, the drain on close, and an eighteen-case refusal matrix in which
every balance, the signer sequence and the state root must be unchanged.
The store-path tests that emitted more than one transfer set through a
single module context now initialise a fresh context per emission, which
is the lifecycle the kernel gives one activity; every existing assertion
in those tests is preserved verbatim.

**No genesis registration row was added.** The budget row already exists
in `module_table[]` in `src/protocol/lxp_genesis.c` with
`LXP_GENESIS_MODULE_GATE_ENABLE_FLAG` and `enabled` false;
`module_table_validate` refuses a duplicate module id. The enable flag
was left false, because flipping it changes the genesis state root and
every snapshot fixture derived from it.

## Gates

| native build | `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/native-build.log` | `fe5a89c71050e02db12c67b3b8286c5b44f2a80a` |
| aggregate test | `make CC=gcc test` | 0 | `qual-logs/make-test.log` | `fe5a89c71050e02db12c67b3b8286c5b44f2a80a` |
| budget targets | `make CC=gcc test-budget-create test-budget-period test-budget-spend test-budget-delegate test-budget-revoke` | 0 | `qual-logs/budget-targets.log` | `fe5a89c71050e02db12c67b3b8286c5b44f2a80a` |
| ledger check | `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `fe5a89c71050e02db12c67b3b8286c5b44f2a80a` |

The five budget targets also run `tools/lxp_check_sole_writer.sh`, which
holds `src/ledger/lxp_apply.c` as the sole writer of `->balance =` and
forbids `src/modules` from touching `lxp_apply_transfer`,
`lxp_balance_apply_leg` or `lxp_balance_restore_snapshot`.

## Observations added

`spec/layerx-beta/qualification.kvx`, task 6.1, ids 2900 to 2906:

- **2900** the pre-existing four-target failure and its
  `source_authorities` root cause.
- **2901** lazy period rollover for key-value records, bounded by
  `LXP_MODULE_MAX_STAGED_WRITES`.
- **2902** the one-transfer-set-per-context rule in `emit_transfer_set`
  and the context lifecycle it requires of the store-path tests.
- **2903** hosting the dispatch suite inside `test_budget_create.c`
  because a new test binary needs a Makefile rule outside this change.
- **2904** amend and the delegate ordinals move no value, so no interface
  advances the signer sequence for them; bounded by the new exact
  sequence check and by the per-record duplicate refusals.
- **2905** the store-path drain reduces `per_period_limit` to
  `spent_this_period`, which `lx_budget_record_validate` refuses to
  encode when the period has no spend; both sides left as written.
- **2906** the genesis row already exists and the enable flag was left
  false.


paxlabs-inc:
Rebased mechanically onto main df8365f2a: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 9167bad24: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head fe5a89c71 (no file overlap with main since the merge base).

PR #222, opened 2026-09-10
Dispatch every escrow ordinal over module-scoped kernel state

paxlabs-inc:
## The issue

The escrow module advertised seven activity types and could execute none of
them. It exposed no `lxp_module_iface`, so no ordinal reached the kernel:
there was no `decode`, no `validate`, no `execute` and no `state_root`, and
`lxp_kernel_module_for_activity` could not resolve an escrow activity at all.
Every hold lived in an `lx_escrow_store` array that the caller passed in
through the request, so hold state never reached module state, never survived
a commit and never contributed to the state root. Capture and release against
an expired hold mutated the hold, moved balances and wrote an idempotency
record before refusing, and no transition emitted an event.

## What changed

**A real module interface.** `src/modules/escrow/lx_escrow_module.c` implements
`decode`, `validate`, `execute`, `genesis`, `epoch_begin`, `epoch_end` and
`state_root` and dispatches all seven advertised ordinals: open, capture,
partial capture, release, timeout, dispute open and dispute resolve. Payloads
are decoded from fixed canonical layouts with typed refusals for length and
shape, the accounts and the asset of each activity are resolved from the hold
record through the module runtime the kernel binds (the same mechanism the
asset and programs modules use), and every settled transition emits one
sixty-seven byte event body carrying the escrow id, the resulting state and
the captured and locked amounts. `state_root` reports the module subtree root.

**Canonical KV state.** `src/modules/escrow/lx_escrow_state.c` holds the record
and economic-result codecs and the module-scoped KV layer: holds are written
under a hold prefix and the projection of every settled activity under a
result prefix, both through `lxp_ctx_kv_get/put/iter`, so state survives a
commit, enumerates in canonical key order and contributes to the state root.
An idempotent replay rebuilds the receipt from the stored result rather than
persisting a receipt structure. The caller-owned `lx_escrow_store` and the
`store` member of every escrow request are gone.

**Ledger-correct settlement.** Each settlement builds its transfer set with a
single source authority over the escrow account, the escrow origin module id
and the escrow debit authority kind, which is what `custody_spend_check` and
`lxp_kernel_apply_transfer_set` require; without them the kernel refused every
escrow transfer set. The two-leg dispute split debits one account, so the
sequence advances once. Post-transfer state writes are reserved with
`lxp_module_staged_reserve` before the transfer and released after it, so no
settlement can move funds and then fail to record them.

**Expiry semantics.** A capture or release against a hold whose expiry has
passed is refused with `LXP_ERR_HOLD_EXPIRED` before any hold state, ledger
balance or idempotency record is touched. The epoch sweep in
`lx_escrow_epoch_begin` is the only writer of the timed-out transition: it
collects expired holds in canonical order, moves the remaining balance of each
in one transfer set, writes the timed-out record and its receipt, emits the
timeout event, and refuses the whole epoch with `LXP_ERR_ARENA_EXHAUSTED`
rather than settling a prefix when more holds expire than one epoch can carry.

**Tests.** The five escrow tests drive the ordinals through the module
interface rather than calling the executors directly: interface metadata and
activity resolution, record and result codec round trips and rejections,
transfer legs, reasons, authority kinds and balances, event bodies, replay
receipts that must be byte-identical and must not move funds twice, and the
refusals for duplicate open, wrong principal, over-capture, capture equal to
the remaining balance, missing hold, expired hold with state, balances and
idempotency record all unchanged, disputed hold, closed dispute window,
premature timeout, unbound runtime, unauthorized escrow spend and sweep
overflow. A sixty-four iteration property loop drives open followed by a
random settlement path and asserts conservation and the hold invariant after
every step, and the timeout test compares the state root of an immediate sweep
against a delayed one.

The genesis module table row for this module already exists and is gated by
its enable flag, which stays clear; no row was added and no flag was flipped.

## Gates

| command | exit | log | commit |
| --- | --- | --- | --- |
| `make CC=gcc test` | 0 | `qual-logs/make-test.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |
| `make CC=gcc test-escrow-open test-escrow-capture test-escrow-timeout test-escrow-invariants` | 0 | `qual-logs/escrow-module-tests.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |
| `make CC=gcc EXTRA_LDFLAGS="programs/target/debug/liblayerx_programs_sandbox.a -ldl -lm" test-escrow-dispute` | 0 | `qual-logs/escrow-dispute-test.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/native-binaries.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |
| `make CC=gcc test-service-dispute` | 2 | `qual-logs/service-dispute-test.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |
| `make CC=gcc test-genesis-bootstrap` | 2 | `qual-logs/genesis-tests.log` | `5faf1a0fd1418bcf6f0e391b1e2c0fc2699f0e0f` |

The last two are recorded, not claimed. `test-escrow-dispute` needs the
`EXTRA_LDFLAGS` hook because its Makefile rule links only the testing archive;
the rule is unchanged and the defect is recorded.

## Observations added

`spec/layerx-beta/qualification.kvx`, all under task 6.11:

- `observation.6.11.2800` (blocker): the `test_escrow_dispute` and
  `test_stream_lifecycle` link rules omit the programs runtime archive, its
  build prerequisite and `-ldl -lm`, so neither links on a clean checkout.
- `observation.6.11.2801` (blocker): `tests/modules/test_service_dispute.c`
  still uses the removed caller-owned store; the four-line adaptation is
  written out for the owner of that path.
- `observation.6.11.2802` (blocker): no process binds an escrow module runtime,
  so escrow activities submitted to a node are refused before reaching the
  module.
- `observation.6.11.2803` (note): the genesis enable flag for this module stays
  clear and the row is not duplicated.
- `observation.6.11.2804` (note): the expiry, sweep-capacity and one
  transfer-set-per-context decisions taken under the row default, including the
  effect-buffer coupling that any module re-initialising a context must honour.
- `observation.6.11.2805` (blocker): `test-genesis-bootstrap` requires the asset
  module to refuse ordinal 9 while the module advertises it; reproduced on an
  unmodified checkout of main.


paxlabs-inc:
Rebased mechanically onto main 2fd794be1: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head f877891f3: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 5faf1a0fd (no file overlap with main since the merge base).

PR #223, opened 2026-09-10
Persist gateway invoice and idempotency state in metered key/value state

paxlabs-inc:
## Issue

Gateway bookkeeping was bounded in memory. `lxp_gateway_invoice_registry` held settled invoices in a fixed `lxp_gateway_invoice_record records[256]` array (`LXP_GATEWAY_INVOICE_CAPACITY`, `include/layerx/lxp_gateway.h:19` on the base revision), so the 257th invoice was refused with `LXP_ERR_ARENA_EXHAUSTED`. Duplicate detection leaned on `lxp_send_store`, whose `LXP_SEND_STORE_CAPACITY` is 64 (`include/layerx/lxp_ledger.h:76`), so the gateway forgot the 65th distinct idempotency key. Neither structure was committed anywhere: rollback was a `memset` over the two arrays (`src/network/lxp_gateway_receive.c:46-73` on the base revision), so a settlement left no state behind and a refusal could not be checked against a state root.

## What changed and why

**Canonical key/value state for invoice and idempotency records.** `src/network/lxp_gateway_kv.c` adds a growable, sorted key/value store owned by the gateway. Keys are namespaced `"gwinv:" || invoice_id || idempotency_key` (70 bytes) and `"gwidem:" || domain || idempotency_key` (40 bytes, domain 1 for send and 2 for receive so the two windows stay separate), ordered by `memcmp` over the common prefix and then shorter first. The commitment follows `src/state/lxp_state_proof.c` exactly: `LXP_DOMAIN_STATE_LEAF` leaves over u32 big-endian key and value lengths, `LXP_DOMAIN_STATE_NODE` internal nodes, an odd node paired with itself, and the leaf domain over no bytes as the empty root.

**Limits only by gas.** Every write charges its net byte delta to `lxp_meter_charge_storage`, so invoice and idempotency counts are bounded by the settlement storage ceiling rather than by an array length. `lxp_gateway_settlement_context` and `lxp_gateway_receive_context` carry the meter; a NULL meter leaves the store unmetered as before.

**Rollback.** Each settlement takes an undo mark before it opens the balance journal and saves `lxp_meter_ctx` by value, because `lxp_meter_charge_storage` mutates the meter even when it refuses. The abort helper rolls the store back to that mark, restores the meter and the idempotency window, and only then reports the original refusal. A refusal at any of the six fault-injection boundaries now leaves balances, the send store, the invoice count, the key/value count, the stored byte total and the state root exactly as they were.

**The 64-entry window.** `lxp_send_store` becomes a write-through window over the persisted history. The gateway reads the persisted history immediately before delegating to `lxp_send_execute` or `lxp_receive_execute`, which is the exact point where the ledger performs its own scan, so refusal ordering is unchanged: an activity-hash match still answers `LXP_ERR_SEQUENCE_REUSED` and an idempotency-key match still answers `LXP_ERR_IDEMPOTENT_REPLAY` with the stored projection and `replayed` set. `encode_common` and `write_receive_core` both cover `idempotency_key`, so activity-hash equality implies key equality and one entry per key holding `activity_hash || projection` reproduces the per-record scan. When the window is full it is mirrored into the store and emptied, with an exact backup restored on abort. The ledger refusals and the 64-entry array are untouched, so duplicate detection is strengthened from the last 64 records to unbounded history without weakening any check.

**Record encoding.** Receipts settled here carry `module_id` 0 because `lxp_ledger_receipt_build` never assigns one, and `lxp_receipt_decode` refuses that at `src/protocol/lxp_receipt.c:792-795`, so the persisted record uses an explicit gateway encoding in the style of `lx_perps_market_encode`: fixed big-endian fields, a length-prefixed effect list, refusal of any receipt carrying a program outcome, and a decode-and-compare against the original receipt before the write is admitted. The mismatch is recorded, not repaired.

**Accessors and tests.** `lxp_gateway_invoice_count`, `lxp_gateway_state_root` and `lxp_gateway_stored_bytes` expose the committed state under the existing registry lifecycle handshake. `tests/network/test_gateway_kv.c` covers the store primitives (order independence, overwrite and create rollback restoring both root and byte total) and settles 300 invoices through one meter, crossing both former limits, then checks replay of the first settlement, sequence reuse, a foreign idempotency key under a fresh invoice, and full rollback when the storage ceiling refuses. The two existing gateway tests no longer reach into the removed array and assert against the key/value count, byte total and state root instead.

## Gates

|---|---|---|---|
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/build.log` | fa4f09f339e93fedbadb7a482c52e634a1b259ec |
| `make CC=gcc test` | 0 | `qual-logs/test.log` | fa4f09f339e93fedbadb7a482c52e634a1b259ec |
| `make CC=gcc test-gateway-send test-gateway-receive test-gateway-kv` | 0 | `qual-logs/test-gateway.log` | fa4f09f339e93fedbadb7a482c52e634a1b259ec |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | fa4f09f339e93fedbadb7a482c52e634a1b259ec |

`make CC=gcc test` does not reach any gateway target, so the three gateway tests were run as a separate gate. `test-gateway-kv` was added to `core-test-all` alongside the existing gateway targets.

## Observations added

- `[observation.3.1.3600]` — why the state is a gateway-owned canonical key/value store rather than kernel module key/value, whose bounds (`include/layerx/lxp_kernel.h:17-20`) are smaller than the state being persisted.
- `[observation.3.1.3601]` — `lxp_ledger_receipt_build` never sets `module_id` or `module_version`, so a settled gateway receipt is encodable but not decodable by `lxp_receipt_decode`.
- `[observation.3.1.3602]` — the ledger-level 64-key refusals in `lxp_send_execute` and `lxp_receive_execute` remain for any caller that does not go through the gateway.
- `[observation.3.1.3603]` — window compaction clears records a non-gateway writer placed in the same store; their duplicate detection is preserved through the persisted history.
- `[observation.3.1.3604]` — the gateway tests and the Makefile were changed alongside the owned sources, and no assertion was relaxed.

The rebase onto `origin/main` consumed the closing `severity` line of `[observation.2.1.3504]`; it was restored before the ledger gate was run, and `git diff origin/main HEAD -- spec/layerx-beta/qualification.kvx` deletes no lines.


paxlabs-inc:
Rebased mechanically onto main fb5cec53f: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 84056b3b7: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head fa4f09f33 (no file overlap with main since the merge base).

PR #224, opened 2026-09-10
Give the service module real dispatch, canonical state and events

paxlabs-inc:
## The issue

`src/modules/service` advertised thirteen activity types and its `execute`
entry point returned `LXP_OK` for every one of them without emitting an
effect or writing a byte of state. An offer, an agreement, a delivery or a
dispute left no trace: no event, no canonical state, nothing a receipt or a
state root could bind. `spec/layerx-protocol/design.md` section 12.5 fixes
the state model and the thirteen activities; none of it was implemented.

## What changed and why

**Dispatch.** A typed payload decoder covers all thirteen ordinals and
refuses truncated, wrong-version, trailing-byte, over-length and
non-canonical input before validation runs. `module_validate` re-checks the
module id and the ordinal against the activity type, refuses a zero
principal and charges gas. `module_execute` requires the effect buffer the
context is bound to, runs the ordinal's handler and then audits the effects
it produced.

**Canonical state.** Every activity reads and writes module key-value state
through `lxp_ctx_kv_get/put` under eight versioned 8-byte prefixes —
`offer::1`, `agree::1`, `commit:1`, `execut:1`, `delivr:1`, `delitm:1`,
`disput:1`, `progrs:1` — with fixed-width big-endian encoders and decoders
for each record. A delivery of sixteen deliverables would exceed the
1024-byte module value cap, so a delivery is stored as a 138-byte header
plus one 73-byte record per deliverable rather than lowering
`LX_SERVICE_MAX_DELIVERABLES`.

**Events.** Every successful activity emits exactly one event whose 73-byte
body is a primary identifier, a secondary identifier, a status code and the
global sequence. No service activity emits a transfer: `design.md` states
"Transfers emitted: none", and `lx_service_effect_audit` runs after every
execution and refuses at `LXP_FATAL_INVARIANT` if any effect carries
`LXP_EFFECT_TRANSFER` or the monetary flag, and at
`LXP_ERR_UNKNOWN_ACTIVITY` for an activity type the interface never
declared.

**Lifecycle.** Agreement formation is a two-step handshake:
`agreement_propose` is buyer-authorised and creates the agreement in a new
`LX_SERVICE_AGREEMENT_PROPOSED` state carrying the offer's delivery
deadline, acceptance window and dispute window with overflow refusals;
`agreement_accept` is provider-authorised, re-checks the offer for
withdrawal, prior acceptance and expiry, and moves the agreement to
`FORMED`. Then follow task commitments, progress reports with a monotonic
high-water mark and a `METER_REGRESSION` refusal, signed tool execution
attestations verified against an attestor grant and the activity
identifier, deliveries checked against the offer's deliverable
specification and availability reference, acceptance and rejection with
contested deliverable hashes, and dispute opening and resolution — each
with its own state transition, authority check and window refusal. The
default outcome an offer declares is applied by a sweep over agreements
whose acceptance window has closed.

**Hash sets.** Duplicate contested and evidence hashes are now refused as
non-canonical rather than stored, so two payloads cannot encode the same
set.

**Tests.** The six existing service tests were rewritten against the
context and key-value API. They drive the module through
`lxp_kernel_dispatch` with real Ed25519 attestation signing and assert the
stored record lengths and bytes, the event bodies, replay refusals,
authority refusals, deadline and window refusals, the deliverable and
progress orderings, the interface shape, the state root changing across a
commit, and that a counting transfer applier installed on the kernel is
never reached across the whole lifecycle.

## Registration row

Not added. Lane M0's genesis module table is already on `main`
(`src/protocol/lxp_genesis.c:39-58`) and already carries the service row
with `LXP_GENESIS_MODULE_GATE_ENABLE_FLAG` and `enabled = false`. A second
row would be refused as a duplicate module id, and flipping the enable flag
would change the genesis state root and every snapshot fixture, so the
module stays default-disabled and is qualified through direct registration.

## Gates

| --- | --- | --- | --- |
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/build.log` | `dec71cfe3` |
| `make CC=gcc test` | 0 | `qual-logs/make-test.log` | `dec71cfe3` |
| `make CC=gcc test-service-offer test-service-commit test-service-attest test-service-deliver test-service-acceptance test-service-dispute` | 0 | `qual-logs/make-test-service.log` | `dec71cfe3` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `dec71cfe3` |

The service targets include `tools/lxp_check_sole_writer.sh`, which passes:
no service source assigns a balance and no service test links the ledger
transfer entry points.

## Observations added

`spec/layerx-beta/qualification.kvx`, six blocks under `task = "3.3"`:

- `observation.3.3.3100` (note) — the propose/accept authority split and the
  new `PROPOSED` state, and why a single-step form was rejected.
- `observation.3.3.3101` (note) — the module advertises thirteen ordinals
  and none is named settle or cancel; `accept`/`reject` and
  `offer_withdraw`/`commit_abandon` are those transitions, and no ordinal
  emits a value leg.
- `observation.3.3.3102` (blocker) — the kernel null-checks `epoch_begin`
  and `epoch_end` at `src/protocol/lxp_kernel.c:991` but never calls either,
  so the implemented default-acceptance sweep has no production caller. The
  same gap affects escrow and budget.
- `observation.3.3.3103` (note) — a full delivery exceeds
  `LXP_MODULE_MAX_VALUE_BYTES`, so the record is split rather than the
  deliverable bound lowered.
- `observation.3.3.3104` (blocker) — the aggregate `make test` target does
  not depend on any module test target, so a service regression cannot fail
  it; both `make CC=gcc test` and the six service targets were run.
- `observation.3.3.3105` (note) — the previous dispute test's escrow
  coupling used a store API that is being replaced; the "no value moved"
  invariant is now proven inside the service module with a counting transfer
  applier.


paxlabs-inc:
Rebased mechanically onto main a5da5bb77: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 653660d75: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head dec71cfe3 (no file overlap with main since the merge base).

PR #227, opened 2026-09-10
Run the go/no-go report check from the beta contract check

paxlabs-inc:
## Issue

The go/no-go report generator (`tools/ci/beta-report.sh`) landed with no wiring:
no make target ran it, and nothing ran its `--check` mode against the contract,
so task 5.4's `verify_cmd` `make beta-report beta-ledger-check beta-contract-check`
could not run as written. Wiring the check in then surfaced the pre-existing
`make beta-contract-check` failure recorded as `[observation.5.4.5403]`:
the derived surfaces `hosted-agentd` and `middleware-examples` had no row in
the contract's surface table.

## What changed and why

**`Makefile`** — `beta-report` is declared on the `.PHONY` line beside
`beta-ledger-check` and `beta-contract-check` and runs `tools/ci/beta-report.sh`.

**`tools/ci/beta-contract-check.sh`** — normalises its contract path to an
absolute path, reduces it to a root-relative argument so the report renders with
the same contract identity string it carries on disk, and runs
`tools/ci/beta-report.sh --check --contract <contract>` after its own python
check. Both checks always run and both violation lists print; the exit status is
the contract status when that is non-zero and the report status otherwise. The
usage text documents the added check and the exit rule.

**`platform/docs/content/beta.md`** — the two derived surfaces missing from
`Surfaces and journeys` are added with their real class, required rung and
source:

| Surface | Class | Required rung | Reached rung | Source |
| hosted-agentd | hosted | deployment_proven | source_present | platform/hosted/agentd |
| middleware-examples | functional | runtime_proven | source_present | platform/middleware/examples |

`source_present` is what the evidence supports: no gate record exists on any
release-candidate revision and `tools/ci/beta-report.sh` records
`source_present` for every surface. Neither derivation in the check was removed,
narrowed or given a new exclusion — the check reads exactly as it did. The
paragraph describing what the check covers now names the report check too.

**`spec/layerx-beta/report.md`** — re-rendered by the script only. It carries 88
surface rows and lists both new surfaces under *Surfaces no evidence-map target
names*, with their reachable-rung ceiling at `source_present`, so the gate-set
gap stays visible rather than being papered over.

## Gates

| --- | --- | --- | --- |
| `make beta-report` | 0 | `qual-logs/beta-report.log` | `853f259ad8c73a0f72519af67fb0f6fb098b351a` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `853f259ad8c73a0f72519af67fb0f6fb098b351a` |
| `make beta-contract-check` | 0 | `qual-logs/beta-contract-check.log` | `853f259ad8c73a0f72519af67fb0f6fb098b351a` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `853f259ad8c73a0f72519af67fb0f6fb098b351a` |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | `853f259ad8c73a0f72519af67fb0f6fb098b351a` |

Propagation was exercised in both directions before the gate run:

| Perturbation | `make beta-contract-check` | Message |
| --- | --- | --- |
| line appended to `spec/layerx-beta/report.md` | 1 | `beta-report: spec/layerx-beta/report.md is not the report this evidence renders` |
| `middleware-examples` row deleted from the contract | 1 | contract violation and report violation both listed |
| unmodified tree | 0 | both summaries printed |

`tools/ci/beta-contract-check.sh --help` still exits 0 and an unknown option
still exits 2; a relative `--contract` given from a different working directory
resolves correctly and exits 0.

## Observations added

`[observation.5.4.7800]` (severity `note`) in `spec/layerx-beta/qualification.kvx`:
records that this change closes the request in `[observation.5.4.5404]` and the
blocker in `[observation.5.4.5403]`, the three propagation runs, the contents of
each new surface's directory that the journey text describes, and the fact that
neither new surface is named in the `tools/ci/beta-report.sh` evidence map, so no
gate command in the repository can raise either above `source_present` today.
That evidence-map gap is outside this change's scope and is left recorded, not
worked around.


paxlabs-inc:
Rebased mechanically onto main 9756b136f: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head bed2bb040: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 853f259ad (no file overlap with main since the merge base).

PR #230, opened 2026-09-10
Require durable autobahn consensus state instead of silently running in memory

paxlabs-inc:
## Issue

Autobahn consensus persistence was optional and its absence was silent.
`persist.NewPersister` installs `newNoOpPersister` whenever it is handed no state
dir, and that persister reports every write as a success while discarding it. A
validator configured without `persistent_state_dir` therefore kept its whole
consensus safety state — the highest view entered and the votes cast in it — in
memory, and restarted with no record of what it had already voted for, which is
the condition that produces equivocation and slashing.

Neither layer objected. `AutobahnFileConfig.Validate` checked validators,
`max_txs_per_block`, `max_txs_per_second`, `block_interval`, `view_timeout` and
`dial_interval`, and never looked at `persistent_state_dir`; `node/setup.go`'s
`buildGigaConfig` copied the field straight into the consensus config.

## What changed and why

`paxeer-network/consensus/config/autobahn.go`

- `Validate` now refuses an absent `persistent_state_dir` with
  `ErrPersistentStateDirRequired`, a blank one with `ErrPersistentStateDirBlank`,
  and the contradictory pair of a state dir plus the opt-out with
  `ErrPersistenceOptOutConflict`. All three are exported sentinels, so callers and
  tests match them with `errors.Is` instead of on message text.
- New field `unsafe_test_only_disable_persistence`, named for what it costs. It is
  the single remaining way to reach the no-op persister. It defaults to false, so a
  config file written before the field existed, and any config that simply omits
  the state dir, fails closed with a message naming the field the operator must
  set. It is mutually exclusive with `persistent_state_dir` so a config cannot
  state both intents at once.
- New accessor `ConsensusPersistentStateDir`, which returns the dir the consensus
  and data layers must persist to, `None` only under the explicit opt-out, and an
  error in every case that would install the no-op persister.

`paxeer-network/consensus/node/setup.go`

- `buildGigaConfig` no longer reads `fc.PersistentStateDir`. It derives the dir
  through `ConsensusPersistentStateDir`, so the site that constructs the consensus
  config refuses the no-op persister on its own rather than relying on `Validate`
  having run first.
- The comment at the rootify site in `createRouter` is corrected: `None` there now
  means the operator set the test-only opt-out, and every other route to `None` has
  already been refused.

An opt-out is kept rather than removed because throwaway test networks legitimately
run in memory, and the data-layer WAL in `internal/p2p/giga_router.go` piggybacks on
the same option; making it explicit and unsafely named preserves that use while
removing the silent default.

Tests

- `config/autobahn_test.go`: a table over present, absent, blank, whitespace-only,
  opted-out and conflicting state dirs; the accessor's three outcomes; and the JSON
  contract, pinning that a document carrying neither field decodes with the opt-out
  off and is refused, while the opt-out survives a round trip.
- `node/setup_test.go`: each refusal and the accepted opt-out asserted through
  `buildGigaConfig` with `errors.Is`, plus that the configured dir is the dir that
  reaches the consensus config. `defaultFileConfig` now persists to a per-test
  `t.TempDir()`, which is what an operator config must do. The former
  `TestBuildGigaConfig_NonePersistentStateDir`, which asserted that an absent state
  dir is accepted, is the contract this issue changes; it is replaced in place by
  the four tests that assert the new contract, not dropped.
- `internal/autobahn/consensus/persist/persist_test.go`: the no-op persister is
  shown to accept writes, leave the directory empty and recover nothing across a
  restart, against a configured dir that recovers what it wrote — the evidence for
  why the config layer must refuse it.

## Gates

| Gate | Command (from `paxeer-network/consensus`) | Exit | Log | SHA |
| format | `gofmt -l` on the five touched files | 0 (no output) | `qual-logs/gofmt.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| vet | `go vet ./config/... ./node/... ./internal/autobahn/...` | 0 | `qual-logs/go-vet.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| test (focused) | `go test -count=1 ./config/... ./node/... -run 'Autobahn\|BuildGigaConfig\|URLJSON'` | 0 | `qual-logs/go-test-focused.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| test (focused) | `go test -count=1 ./internal/autobahn/consensus/persist/... -run 'NoneDirIsNoOp\|SomeDirSurvivesRestart'` | 0 | `qual-logs/go-test-focused.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| test (aggregate) | `go test ./...` | 1 — pre-existing failures only; see below | `qual-logs/go-test-all.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| ledger | `tools/ci/beta-ledger-check.sh` (repository root) | 0 | `qual-logs/beta-ledger-check.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| report | `tools/ci/beta-report.sh --check` (repository root) | 0 | `qual-logs/beta-report-check.log` | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |
| whitespace | `git diff --check` | 0 | — | `6f38e0416ad727c213ffe8016fe3518f99cf3a40` |

The aggregate `go test ./...` exits 1 on packages that this change cannot reach:
`go list -deps -test` shows that not one of them compiles against `consensus/config`
or `consensus/node`, the only non-test packages touched here, and the failures
reproduce on the unmodified worktree. The suite was run twice, before and after the
rebase; the inventory below covers both runs.

- `internal/autobahn/consensus`, `internal/autobahn/consensus/persist` and
  `internal/autobahn/data` each panic at `autobahn/types/msg.go:167` with
  `autobahn signing key is not bound to a committee`. The panic aborts the test
  binary, which is why the two persist tests added here are shown by a selective run.
- `internal/libs/clist` hangs in `TestGCRandom` until the ten minute test timeout in
  both runs.
- `rpc/jsonrpc/server` fails `TestServeTLS`, which calls `ServeTLS` with the fixture
  pair `test.crt` and `test.key`. `test.crt` is tracked, but `test.key` is matched by
  the `*.key` rule at `.gitignore:64` and so is absent from every checkout; the test
  cannot pass in a fresh clone. It surfaced as `open test.key: no such file or
  directory` in the second run and as a ten minute hang in the first, because the
  test reads the `ServeTLS` error channel for only 100ms before issuing its request.
- `internal/p2p/pex` is flaky between the runs: `TestReactorErrorsOnReceivingTooManyPeers`
  hung to the timeout in the first and the package passed in 0.721s in the second.
- Three persist tests (`TestPersistWriteErrorReturnsError`,
  `TestPersistWriteErrorDoesNotAdvanceSeq`, `TestLoadPersistedOSErrorPropagates`)
  fail whenever the suite runs as uid 0: they revoke write permission with `chmod`
  and require an error, but the effective user bypasses Unix permission checks.
  `persist.go` already documents that limitation for its own writability probe.

No assertion was relaxed, no bound loosened, no check disabled and no test skipped
to reach any of these results.

## Observations added

`spec/layerx-beta/qualification.kvx`, task `6.6`:

- `6.6.4000` (note) — the no-op persister path, the new refusals, and why the opt-out
  exists and is named as it is.
- `6.6.4001` (blocker) — `cmd/tendermint/commands/gen_autobahn_config.go` is outside
  this lane's owned paths, and its `--persistent-state-dir=` (empty) form now emits a
  config file the node refuses at load. The default `data/autobahn` path, which
  `paxeer-network/docker/localnode/scripts/step4_config_override.sh` uses, is
  unaffected. The generator needs a matching explicit flag.
- `6.6.4002` (blocker) — the three committee-binding panics, with the evidence that
  they predate this change.
- `6.6.4003` (blocker) — the `clist` hang, the missing `test.key` fixture behind the
  `*.key` ignore rule, the flaky `pex` test, and the three uid-0 `chmod` tests.

`spec/layerx-beta/report.md` is regenerated by `tools/ci/beta-report.sh` and committed
with the ledger change; it was re-rendered again after the rebase.


paxlabs-inc:
Rebased mechanically onto main b025b77ea: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 141167cc8: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 6f38e0416 (no file overlap with main since the merge base).

PR #231, opened 2026-09-10
Serve public self-service registration through lx_register

paxlabs-inc:
## Issue

Registration was the missing first step of the public product path: the gateway
served fifteen public JSON-RPC methods but none of them created an identity, the
CLI had no `register` command, and the only account-opening command posts to a
path the hosted core lists as an unavailable capability. A newcomer holding a
local signing key had no served way to become a principal.

## What changed and why

**`lx_register`, a new public JSON-RPC method** (`platform/hosted/gateway/src/rpc_register.rs`).
It takes two lowercase-hex arguments — a 32-byte Ed25519 signer public key and a
64-byte signature — and creates one identity principal in the `beta` tenant
through the identity service's `POST /v1/principals`.

The subject is **not** a parameter. It is derived:

```
sub     = "beta." + hex(SHA256(len-prefixed "layerx-register-subject-v1", tenant, key))[..32]
binding =           SHA256(len-prefixed "layerx-register-binding-v1",  tenant, key)
```

and the caller must present an Ed25519 signature over `binding` before the
gateway spends its provisioning credential. This matters because
`Store::put_principal` (`platform/hosted/identity/src/store.rs:107`) inserts by
`sub` and therefore *overwrites*: a caller-chosen subject would let anyone rebind
another principal's signer keys. With a derived subject and proof of possession a
caller can only ever write the principal its own key already owns, and repeating
the call is idempotent. No session is minted — public keys are public, so issuing
a bearer session for one would reintroduce exactly the takeover this closes.

Refusals are typed and carry a machine-readable `error.data.code`:
`-32602` malformed arguments, `-32002` `registration_proof_invalid`,
`-32005` `registration_rate_limit`, `-32001` `registration_not_configured` /
`identity_unavailable`. A 200 from identity that does not confirm the tenant, the
derived subject and exactly the presented key is `-32603`, not a success.

**Rate limit.** Self-registration is bounded by a per-process minute window of 30
registrations, split into a pure `consume(window, minute)` that the tests drive
and a `consume_registration()` that locks the static, so a shared counter later
replaces one function.

**Credential.** `POST /v1/principals` is accepted only for the *provisioning*
service token, which the gateway pod does not currently hold. The token is read
from an optional `LAYERX_GATEWAY_IDENTITY_PROVISIONING_TOKEN_FILE`; when it is
unset the method answers `-32001 registration_not_configured` rather than
pretending. The deployment wiring is outside this branch's paths and is recorded
as an observation.

**CLI** (`platform/cli/src/register.rs`): `layerx register [--key <name>]` reads
the seed from credential storage, checks the derived public key against the one
recorded for that key, signs the binding, posts `/rpc`, and re-derives the
subject locally so a response that names anything else is refused.

**SDK** (`agent/crates/layerx-sdk/src/register.rs`): `RpcClient::register` plus a
`Registration` type whose decoder refuses any principal whose tenant, subject or
single allowed signer key is not what the SDK derives for the key it registered.

**Schema**: one `lx_register` entry in `platform/hosted/gateway/openrpc.json`,
placed first so sibling public-method branches append after it.

The gateway, the CLI and the SDK each pin the same binding and subject vectors in
their own tests, so the three derivations cannot drift apart silently.

## Gates

All gates ran on `5410ac5251052f093181f0ce8bb8b32989b275d7`.

| --- | --- | --- | --- |
| gateway tests | `cargo test --manifest-path platform/Cargo.toml -p layerx-platform-gateway` | 0 | `qual-logs/gateway-test.log` |
| CLI tests | `cargo test --manifest-path platform/Cargo.toml -p layerx-platform-cli` | 0 | `qual-logs/cli-test.log` |
| SDK tests | `cargo test --manifest-path agent/Cargo.toml -p layerx-sdk` | 0 | `qual-logs/sdk-test.log` |
| SDK lints | `cargo clippy --manifest-path agent/Cargo.toml -p layerx-sdk --all-targets -- -D warnings` | 0 | `qual-logs/sdk-clippy-isolated.log` |
| ledger | `make beta-ledger-check` | 0 | `qual-logs/ledger-check.log` |
| report | `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` |

The gateway and CLI gates were run together as one `cargo test` invocation over both
packages; `qual-logs/platform-tests.log` is that run and both gate logs are copies of it.
It reports `42 passed` for the gateway binary (including the eight `rpc_register` tests) and
`36 passed` for the CLI binary (including the four `register` tests), with every other target
green and nothing ignored or filtered.

Two environment notes, neither a property of this branch:

- `cargo test -p layerx-platform-cli` **deadlocks** whenever `CARGO_TARGET_DIR` is exported.
  `layerx-platform-emulator`'s build script runs `make build`, whose `programs-build` recipe
  runs `cd programs && cargo build`; that inner cargo inherits the variable and blocks forever
  on the build-directory lock its own parent holds. The gate above was run with
  `PROGRAMS_CARGO='env -u CARGO_TARGET_DIR cargo'`. Recorded as `observation.6.5.4205` with the
  one-line repair.
- SDK lints were run in a private target directory. In the shared one the
  `layerx-agent-api` build script failed on a generated-source drift assertion against a stale
  cached build-script binary; the same crate cold-builds clean (exit 0) from an untouched
  checkout of its files, which are byte-identical to `main`.

## Observations added

`spec/layerx-beta/qualification.kvx`, all under `task = "6.5"` and inside this lane's range:

| Id | Severity | Subject |
| --- | --- | --- |
| `6.5.4200` | blocker | The gateway pod holds only its own identity token; `POST /v1/principals` needs the provisioning token, so the deployment must mount it. |
| `6.5.4201` | note | `put_principal` overwrites by `sub`; why the subject is derived and possession proven instead. |
| `6.5.4202` | note | The registration budget is per-process, not cluster-wide, and how it is factored for a shared counter. |
| `6.5.4203` | blocker | `layerx account create` still posts to `/v1/accounts`, which the hosted core lists as an unavailable capability. |
| `6.5.4204` | note | The OpenRPC method-count assertion is the one place the three public-method branches meet. |
| `6.5.4205` | blocker | `programs-build` deadlocks any cargo gate run with an exported `CARGO_TARGET_DIR`. |

`spec/layerx-beta/report.md` was regenerated by `tools/ci/beta-report.sh` in the same commit and
never edited by hand.


paxlabs-inc:
Rebased mechanically onto main bb328afd8: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 40bd82b91: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 5410ac525 (no file overlap with main since the merge base).

PR #233, opened 2026-09-10
Refuse a terminal account sequence instead of wrapping it

paxlabs-inc:
## The issue

`lx_account.next_sequence` was advanced unconditionally at every site that
consumes an account sequence. An account that reaches `UINT64_MAX` therefore
wraps back to zero on the next activity, and every sequence it has already
consumed becomes replayable — the replay protection the counter exists to
provide disappears exactly when the account has used it the most.

Four sites were affected: `lxp_apply_transfer` and `lxp_apply_transfer_set` in
the ledger, and the `lxp_send_validate` / `verify_receiver` sequence
comparisons, which accepted `UINT64_MAX` as a perfectly ordinary next sequence.

## What changed

- `include/layerx/lxp_result.h` gains `LXP_ERR_SEQUENCE_EXHAUSTED` as `-307`,
  appended at the end of the sequencing domain. No existing value moved.
- `src/ledger/lxp_apply.c` defines `lxp_sequence_terminal_check`, which refuses
  when either the debit source `leg->from` or the account named by
  `context->sequence_account` stands at `UINT64_MAX`. `lxp_precondition_check`
  consults it before any balance moves; `lxp_apply_transfer` consults it again
  before `lxp_balance_apply_leg`.
- `src/ledger/lxp_journal.c` consults it immediately before the transfer-set
  increment. A refusal rolls the journal back and reports the failed leg with a
  zero leg count, so a set that has already mutated earlier legs leaves no
  partial state behind.
- `src/ledger/lxp_send.c` and `src/ledger/lxp_receive.c` answer the new code
  ahead of their `LXP_ERR_SEQUENCE_MISMATCH` comparison, so an exhausted
  account is never reported as a caller that guessed the wrong sequence.
- `src/ledger/lxp_ledger_internal.h` declares the predicate. That private
  header is included by `src/ledger/lxp_journal.c` alone and already carries
  the same pattern for `lxp_balance_restore_snapshot`.

The rule chosen is that the terminal value is unusable: the account keeps its
balance and can still be credited as the recipient of a leg, but it cannot
debit and cannot serve as the sequence account. The occupancy carve-out is
untouched — a leg carrying `LXP_AUTH_OCCUPANCY_RESPONSIBILITY` consumes no
sequence and is unaffected until its payer itself goes terminal.

Because `lxp_terminal_rejection_applies` classifies by result domain and
returns true for `LXP_RESULT_DOMAIN_SEQUENCING`, the new code is published as a
terminal rejection with no change to that classifier.

## Tests

`tests/ledger/test_apply_transfer.c`, `test_transfer_set.c`, `test_send.c` and
`test_receive.c` cover: a refused debit with balances unchanged and no wrap; a
terminal account still being credited while its own sequence stays at
`UINT64_MAX`; a terminal sequence account refused even when the debit source
could still advance; the occupancy carve-out succeeding without an increment
and then being refused once its payer is terminal; a transfer set refused at
leg 3 after legs 0-2 had already mutated, asserting the rollback restores every
balance; a set refused at leg 0 with no mutation at all; and recovery — the
same send, receive and set succeed once the account is spendable again.

## Gates

| --- | --- | --- | --- |
| `make -j4 CC=gcc test` | 0 | `qual-logs/make-test.log` | `8c2fd3ba5` |
| `make CC=gcc test-ledger-transfer` | 0 | `qual-logs/test-ledger-transfer.log` | `8c2fd3ba5` |
| `make CC=gcc test-ledger-set` | 0 | `qual-logs/test-ledger-set.log` | `8c2fd3ba5` |
| `make CC=gcc test-ledger-send` | 0 | `qual-logs/test-ledger-send.log` | `8c2fd3ba5` |
| `make CC=gcc test-ledger-receive` | 0 | `qual-logs/test-ledger-receive.log` | `8c2fd3ba5` |
| `make CC=gcc test-ledger-receipt` | 0 | `qual-logs/test-ledger-receipt.log` | `8c2fd3ba5` |
| `make CC=gcc test-ledger-accounts` | 0 | `qual-logs/test-ledger-accounts.log` | `8c2fd3ba5` |
| `make CC=gcc beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `8c2fd3ba5` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `8c2fd3ba5` |
| `make -k -j4 CC=gcc` over the seventeen ledger-consumer targets (recorded, not a required gate) | 2 | `qual-logs/ledger-consumers-head.log` | `8c2fd3ba5` |

The recorded batch covers `test-escrow-open`, `test-escrow-capture`,
`test-escrow-timeout`, `test-escrow-invariants`, `test-escrow-dispute`,
`test-budget-create`, `test-budget-spend`, `test-budget-delegate`,
`test-budget-revoke`, `test-stream-open`, `test-stream-settle`,
`test-stream-lifecycle`, `test-service-dispute`, `test-perps-margin`,
`test-perps-funding`, `test-perps-liquidation` and `test-perps-insurance`.
Fifteen of the seventeen build and pass. The two failures are link errors in
`Makefile:496` and `Makefile:612`, which name `$(TEST_LIBRARY)` without
`$(PROGRAMS_RUNTIME_LIB)` and leave the Rust programs-sandbox entry points
undefined; this change adds no symbol to either binary and touches no Makefile
rule.

## Observations added

`spec/layerx-beta/qualification.kvx` gains six blocks under task `6.11`:

| Block | Severity | Subject |
| --- | --- | --- |
| `observation.6.11.2400` | note | why the new code is `-307` at the end of the sequencing domain and why it needs a value of its own |
| `observation.6.11.2401` | note | why the predicate is declared in the private `src/ledger/lxp_ledger_internal.h` rather than duplicated per translation unit |
| `observation.6.11.2402` | suspect | `agent/crates/layerx-types/src/result.rs` mirrors the C result table by hand and has no `-307` variant, and no gate compares the two tables |
| `observation.6.11.2403` | suspect | the two module test link rules that omit `$(PROGRAMS_RUNTIME_LIB)` |
| `observation.6.11.2405` | note | where the refusal is placed, the preserved occupancy carve-out, and the decision that a terminal account keeps its funds and can still be credited |
| `observation.6.11.2406` | note | the measured state of the seventeen ledger-consumer targets on this revision |

`spec/layerx-beta/report.md` is re-rendered by `tools/ci/beta-report.sh` in the
same commit as the ledger change; `--check` exits 0.


paxlabs-inc:
Rebased mechanically onto main 130f4d17e: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 0273a859b: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 8c2fd3ba5 (no file overlap with main since the merge base).

PR #239, opened 2026-09-10
Derive the genesis request length bound from the certificate threshold

paxlabs-inc:
## Issue

`platform/hosted/node/bootstrap.sh` refused the genesis request it had just built whenever the settlement document named more than one guarantor, which is the shipped configuration. The length assertion described a single-guarantor request, so `layerx-genesis-build` was never reached and every harness that bootstraps a node through the script stopped there. Recorded on main as observation `6.11.150`.

## What changed and why

`platform/hosted/node/bootstrap.sh:450` required the LXGB v2 genesis request to measure exactly `395 + metadata` bytes. The request the script assembles carries one 81-byte guarantor record per guarantor (32-byte identifier, 33-byte compressed secp256k1 public key, 16-byte bond), and the number of records comes from `.finality_policy.certificate_threshold` in `contracts/config/checkpoint-settlement.json`, which ships as `2`.

Reading the layout that `cmd/layerx-genesis/lxp_genesis_build_cli.c` `parse_request` decodes, the request is 89 bytes before the guarantor array (magic, envelope version, protocol version, network id, timestamp, parameter count and one 66-byte parameter, guarantor count) and 225 bytes after it (asset id, metering schedule, fee genesis parameters), so its size is `314 + 81 * GUARANTOR_COUNT + metadata`. With the shipped threshold that is 851 bytes against the 770 the script expected — an excess of exactly one guarantor record. `395` is that expression with `GUARANTOR_COUNT` fixed at one; it was the correct expression before the metadata term was introduced. The bound now follows the threshold again.

`agent/crates/layerx-client/tests/availability_daemon.rs` keeps the produced harness evidence and reads the request the bootstrap actually wrote, requiring its envelope version, its single genesis parameter, a guarantor count equal to the certificate threshold read from the settlement document, one 81-byte record per guarantor with a zero bond, a compressed public key prefix and strictly ascending identifiers, one written guarantor private key per record, and the total length the layout implies. Without that, removing the length check entirely would leave the suite green.

With the bound corrected the bootstrap runs to completion against the shipped settlement document: it generates two sorted secp256k1 guarantor keys, emits the 851-byte request, and `layerx-genesis-build` accepts it and prints the network 77 genesis state root, the treasury account and the data and run directories.

## Gates

| --- | --- | --- | --- |
| `bash -n platform/hosted/node/bootstrap.sh` | 0 | `qual-logs/bash-n-bootstrap.log` | `80e171b2a` |
| `rustfmt --edition 2021 --check agent/crates/layerx-client/tests/availability_daemon.rs` | 0 | `qual-logs/fmt-check.log` | `80e171b2a` |
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build build/tests/lxp_test_program_admission build/tests/lxp_test_daemon_finality_authority` | 0 | `qual-logs/native-build.log` | `80e171b2a` |
| `CARGO_TARGET_DIR=/root/lx-shared-target/agent make CC=gcc test-daemon-availability` | 2 | `qual-logs/test-daemon-availability.log` | `80e171b2a` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `80e171b2a` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `80e171b2a` |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | `80e171b2a` |

`test-daemon-availability` does not pass. The length bound this change fixes is no longer the refusal: the script now builds and the check accepts the 851-byte request, and the run stops on two further defects that the bound had been masking, both recorded rather than fixed because neither lies in this change's files. Both were reproduced with the bound corrected and the offending fixture corrected only in the working tree, never committed:

- With the shipped fixture, `layerx-genesis-build` refuses the request because the asset record in `tests/support/lxgb_metadata.py` and `tests/support/lxgb_metadata.rs` no longer satisfies `issuer_valid`. Evidence in `qual-logs/probe-asset-record.log`.
- With the fixture corrected in the working tree, the bootstrap completes, the sequencer and replica start, and `build/tests/lxp_test_program_admission` refuses at its handshake because the daemon advertises LNI 1.5 while the client pins 1.4. Evidence in `qual-logs/proof-fixture-corrected.log`.

## Observations added

| Record | Severity | Subject |
| --- | --- | --- |
| `6.1.5800` | note | The length bound now follows the certificate threshold; supersedes `6.11.150`, with the end-to-end bootstrap evidence and the assertions the test now makes |
| `6.1.5801` | blocker | `tests/support/lxgb_metadata.py` and `.rs` emit a version 3 asset record with custody kind 1, no custody reference and issuer kind 2, which `issuer_valid` refuses; the only shape the builder accepts is issuer kind 2 with custody kind 2 and a non-empty reference. Affects eight harnesses that build genesis through the fixture |
| `6.1.5802` | blocker | `cmd/layerxd/lxp_daemon_lni.c` advertises LNI 1.5 while `tests/daemon/lxp_test_program_admission.c` and the Rust client tests pin 1.4, so the admission handshake refuses a daemon that started correctly |
| `6.1.5803` | note | `bootstrap.sh:202` bounds the genesis metadata by `16384 - 395`, a single-guarantor constant; at the maximum threshold admitted metadata can still carry the request past the builder limit. No current failure |

Observation `6.1.4801`, recorded against this same arithmetic on another branch and not present on this revision, is resolved by this change.

## NEEDS

- `tests/support/lxgb_metadata.py` and `tests/support/lxgb_metadata.rs`: the asset record must be brought back to a shape `lx_asset_record_decode` and `parse_request` both accept, together, since the two fixtures must stay byte-identical and the choice moves the genesis state root every consuming harness builds.
- `cmd/layerxd/lxp_daemon_lni.c` and its clients: the LNI minor version must be reconciled across the daemon, `tests/daemon/lxp_test_program_admission.c` and the Rust client handshake configuration.


paxlabs-inc:
Rebased mechanically onto main ac045fc05: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head a2c5c4b02: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 80e171b2a (no file overlap with main since the merge base).

PR #242, opened 2026-09-10
Meter send and receive history instead of bounding it at sixty-four

paxlabs-inc:
## Issue

`LXP_SEND_STORE_CAPACITY = 64` still drove refusals for every non-gateway ledger
caller. `lxp_send_execute` returned `LXP_ERR_ARENA_EXHAUSTED` once its
`lxp_send_store` array was full and `lxp_receive_execute` did the same for its
idempotency window, so the sixty-fifth distinct idempotency key was refused on
any path that did not go through the gateway. The gateway had already moved its
own duplicate history into metered key/value state, which left the ledger
primitives as the last fixed bound (observation 3.1.3602).

## What changed and why

A shared, gas-metered ledger history in `src/ledger/lxp_send_store.c`:

- `lxp_send_store_lookup` reads the fixed window first, in its original order and
  with its original activity-hash-before-key precedence, then a sorted overflow
  history keyed by idempotency key.
- `lxp_send_store_admit` spills the whole window into that history and empties it
  when the window is full, at exactly the point the refusal used to sit
  (`src/ledger/lxp_send.c`, `src/ledger/lxp_receive.c`), so refusal ordering
  relative to `lxp_send_validate` is unchanged.
- The spill is atomic and metered. It counts the window records the history does
  not already hold, grows the history to fit them, charges that count times
  `LXP_SEND_HISTORY_RECORD_BYTES` (161 = 32 key + 32 activity hash + the 97-byte
  canonical projection the gateway already meters as
  `LXP_GATEWAY_KV_PROJECTION_BYTES`) through `lxp_meter_charge_storage` in one
  call, and only then inserts. A refused charge leaves window and history exactly
  as they were, so `lxp_send_store_total` never counts a record twice and a retry
  never double-charges. History is now bounded only by the execution's storage
  ceiling and the allocator.
- A store carrying no meter cannot spill at all: `lxp_send_store_admit` returns
  `LXP_ERR_ARENA_EXHAUSTED` when `store->meter` is NULL, which is the exact
  refusal the fixed array used to raise. Where there is no budget there is no
  growth, so lifting the bound cannot become an unmetered memory-exhaustion
  vector for an unbudgeted caller. No production caller reaches that refusal on
  this revision: both gateway paths drain their window into key/value state
  before delegating, and the asset module runs a single receive against a
  freshly allocated store.

Replay and window semantics are byte-identical for every receipt an unspilled
window can hold. A key found in the history returns `LXP_ERR_SEQUENCE_REUSED`
for a matching activity hash and `LXP_ERR_IDEMPOTENT_REPLAY` with the stored
projection and `replayed` set otherwise. A keyed binary search is
outcome-identical to the old linear scan because an activity hash covers the
encoded send including its idempotency key and each key appears at most once
across window and history — the same equivalence
`lxp_gateway_idempotency_precheck` already relies on.

The sixty-four-entry array survives as the write-through window because three
gateway files still index it directly and copy the store by value. The history
hangs off a stable pointer added after `count`, so those copies and the gateway's
abort restore stay correct, and history insertion is idempotent for an identical
record. The gateway drains its own window to key/value state before delegating,
so the ledger spill is unreachable from it and no gateway behaviour changes.
Retiring the array is recorded as follow-up work (observation 3.1.7901).

The meter is reached through `lxp_send_store.meter` rather than a new
environment field: both environments are built by positional compound literals in
six places, two of them outside this issue's paths, and the native build refuses
a new trailing field at every one under `-Werror=missing-field-initializers`
(observation 3.1.7902).

## Result-code parity (hand-off from PR 233)

`agent/crates/layerx-types` gains a build script that parses
`LXP_RESULT_CODE_LIST` from `include/layerx/lxp_result.h` and the
`protocol_result_codes!` mirror in `src/result.rs`, and fails the build on any
name, value or position drift. It found the mirror missing two codes, not one:
`SequenceExhausted = -307` (the hand-off) and `ProgramRefused = -736`, which had
drifted out earlier and independently. Both are added in table order. The gate
was exercised against four drift shapes — a deleted mirror entry, a changed
mirror value, an extra mirror entry and a header entry the mirror lacks — and
refuses each one (`qual-logs/parity-gate.log`).

## Tests

`tests/ledger/test_send.c` and `tests/ledger/test_receive.c` drive 67 and 66
distinct idempotency keys through the ledger entry points with no gateway, and
assert the spilled counts, the metered byte total on both the history and the
meter, `LXP_ERR_SEQUENCE_REUSED` for a verbatim re-execute of a spilled send,
`LXP_ERR_IDEMPOTENT_REPLAY` with a byte-identical projection for a different send
reusing a spilled key, and `LXP_ERR_GAS_EXHAUSTED` with unchanged balances, an empty
history and an intact window when the storage ceiling is one byte short of the
spill. A third block drives 65 sends
through a store with no meter and asserts the old `LXP_ERR_ARENA_EXHAUSTED`
refusal with the window full, no history allocated and balances unchanged. No
test pinning 64 as a refusal remained to migrate; the existing suites assert the
window, not the bound.

## Gates

Head `86bd126fc`, rebased onto `8b62a0f4f`. Every gate below ran; the build and
test gates ran at `76d514f85`, whose tree is byte-identical to the head for every
path this branch touches and for every path `make test` compiles. The only
difference between the two revisions outside `spec/layerx-beta` is the upstream
`human-test-hosted-provisioning` target in the `Makefile`, which no prerequisite
of `test` reaches (`git diff 76d514f85 86bd126fc -- include src tests agent` is
empty).

| --- | --- | --- | --- |
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/build.log` | 76d514f85 |
| `make CC=gcc test` | 0 | `qual-logs/make-test.log` | 76d514f85 |
| `make CC=gcc test-gateway-send test-gateway-receive test-gateway-kv` | 0 | `qual-logs/gateway-tests.log` | 76d514f85 |
| `make CC=gcc test-ledger-send test-ledger-receive` | 0 | `qual-logs/ledger-tests.log` | 76d514f85 |
| `CARGO_TARGET_DIR=/root/lx-shared-target/agent cargo test -p layerx-types` | 0 | `qual-logs/cargo-types.log` | 76d514f85 |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | 86bd126fc |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report.log` | 86bd126fc |
| `git diff --check` | 0 | `qual-logs/diff-check.log` | 86bd126fc |
| result-code parity, real crate and four drift shapes (`qual-logs/parity-gate.sh`) | 0 | `qual-logs/parity-gate.log` | 86bd126fc |

`cargo test -p layerx-types` runs the new build script in the loop, so the parity
gate is enforced by a real cargo build and not only by the standalone script.

## Observations added

- `3.1.7900` — closes `3.1.3602`: the bound is lifted for every ledger caller,
  history is metered, and a store with no meter keeps the old refusal.
- `3.1.7901` — the window array survives only because three gateway files index
  it directly; retiring it is gateway-owned.
- `3.1.7902` — the meter lives on the store because the environments cannot gain
  a field under `-Werror=missing-field-initializers`.
- `3.1.7903` — **blocker**: `interop/deploy/mirror/Dockerfile` copies `agent`,
  `human` and `interop` but not `include`, so the new build script cannot find
  the normative table in that image. One-line fix: `COPY include ./include`. No
  silent fallback was added, because a parity gate that skips itself when it
  cannot see the table is not a gate.
- `3.1.7904` — the mirror had drifted by two codes, not one.
- `3.1.7905` — `src/modules/asset/lx_asset_execution.h` releases an
  `lxp_send_store` with a bare `free`; latent today, should become
  `lxp_send_store_release`.


paxlabs-inc:
Rebased mechanically onto main 1960d9ca8: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 63f2e3981: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 86bd126fc (no file overlap with main since the merge base).

PR #243, opened 2026-09-10
Stamp the executing module identity on ledger receipts

paxlabs-inc:
## Issue

`lxp_ledger_receipt_build` (`src/ledger/lxp_receipt.c`) zeroed the receipt and then filled every field except `module_id` and `module_version`. `lxp_receipt_encode` accepted those receipts, but `lxp_receipt_decode` refused them with `LXP_ERR_NON_CANONICAL` because the canonical form requires a non-zero module id and version, and `lxp_receipt_validate_supply` refused a supply-bound ledger receipt outright because it requires `module_id == LXP_MODULE_ASSET`. Every receipt the ledger issued and every receipt `lxp_gateway_send_settle` returned was therefore encodable and returnable over the wire but could not be read back by the protocol codec. Recorded as observation `3.1.3601`, severity blocker.

## What changed

- `lxp_ledger_receipt_build` now sets `module_id` and `module_version` from the executing module context.
- The identity is the asset module. `lxp_ctx_bind_ledger_receipt` refuses any ledger admission whose module is not `LXP_MODULE_ASSET` and whose activity type is not the asset module id composed with the operation, `receipt_bind_ledger_projection` in the kernel does the same, and `lx_asset_module_iface` declares `abi_version` 1.
- Referencing `lx_asset_module_iface` from the ledger translation unit pulls the kernel and programs module objects into every binary that links only `build/liblayerx.a`, leaving ten `layerx_programs_*` symbols undefined. The values are therefore declared as `LXP_LEDGER_RECEIPT_MODULE_ID` and `LXP_LEDGER_RECEIPT_MODULE_VERSION` in `include/layerx/lxp_receipt.h` and tied to the registry constant by a `_Static_assert`, with the test cross-checking both against the real module interface.
- The fix lives in the shared builder, not in `src/network`, so the gateway receipt and the direct ledger receipt continue to agree byte for byte and the equality asserted in `tests/test_gateway_send.c` is unchanged.

### Test coverage

`tests/ledger/test_receipt.c` keeps every existing assertion (balance writer guard, tamper detection, invariant refusal with an unchanged log write offset) and adds:

- the stamped identity, cross-checked against `lx_asset_module_iface()->module_id` and `->abi_version`;
- a decode round trip of the receipt read back from the write-ahead log body, comparing every encoded field, the signature, and `lxp_receipt_verify` against the sequencer public key;
- a re-encode of the decoded receipt byte-compared against the log body;
- the unsigned encoding refused with `LXP_ERR_BAD_SIGNATURE` when a signature is demanded and accepted when it is not;
- a re-signed copy carrying a zero module identity, which must decode as `LXP_ERR_NON_CANONICAL`.

The test now links the programs runtime so it can read the asset module interface, which is why its Makefile rule matches the neighbouring asset test rules.

### Receipt bytes

Receipt bytes and signatures change, because the module identity is covered by the canonical encoding and the digest. No committed artifact pins the old bytes. Decoding every committed receipt artifact with the protocol codec returns `LXP_OK` with `module_id` 1 and `module_version` 1:

| artifact | decode | length | module_id | module_version | operation | supply |
| --- | --- | --- | --- | --- | --- | --- |
| `receipt-positive-v1` | `LXP_OK` | 610 | 1 | 1 | 5 | 0 |
| `receipt-positive-v2` | `LXP_OK` | 610 | 1 | 1 | 5 | 0 |
| `receipt-positive-v3` | `LXP_OK` | 610 | 1 | 1 | 1 | 0 |
| `tests/fixtures/receipt-supply-v2.bin` | `LXP_OK` | 642 | 1 | 1 | 5 | 1 |

Their generators already overwrote the identity before signing, so no fixture or vector required regeneration.

## Gates

| command | exit | log | SHA |
| --- | --- | --- | --- |
| `make CC=gcc test` | 0 | `qual-logs/make-test.log` | `006499938` |
| `make CC=gcc test-gateway-send test-gateway-receive` | 0 | `qual-logs/make-gateway.log` | `006499938` |
| `make CC=gcc test-ledger-receipt` | 0 | `qual-logs/test-ledger-receipt.log` | `006499938` |
| `make CC=gcc test-receipts test-receipt-offline test-gateway test-gateway-kv test-ledger-send test-ledger-receive test-ledger-set test-ledger-transfer test-ledger-accounts test-asset-registry test-asset-transfer test-asset-balance test-tools test-layerxd` | 0 | `qual-logs/make-receipt-suite.log` | `006499938` |
| `make CC=gcc test-replay-golden-local` | 0 | `qual-logs/make-replay-golden.log` | `006499938` |
| `make CC=gcc beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `006499938` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report.log` | `006499938` |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | `006499938` |

## Observations added

| record | severity | subject |
| --- | --- | --- |
| `3.1.8300` | note | supersedes blocker `3.1.3601`: the module identity is now stamped in the ledger builder, receipt bytes and signatures change, no committed artifact required regeneration |
| `3.1.8301` | suspect | `platform/sdk/conformance/fixtures/generate_receipt_fixture.c` cannot run on this revision: `world_init` builds an asset record `issuer_valid` refuses, reproduced with this change reverted |
| `3.1.8302` | note | `parameter_version` is still left unset by the ledger builder while the conformance generators set it to 1 before signing |
| `3.1.8303` | note | the gateway invoice record encoder in `src/network/lxp_gateway_kv.c` exists because ledger receipts could not be decoded, which is no longer true |


paxlabs-inc:
Rebased mechanically onto main 6f952272b: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head d5f0e0a91: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 006499938 (no file overlap with main since the merge base).

PR #244, opened 2026-09-10
Fund a new signer from the public RPC surface through the hosted faucet

paxlabs-inc:
## The issue

The hosted faucet only accepted a claim from a caller that already held an
identity session of its own (`platform/hosted/faucet/src/main.rs:470`), so
nothing on the public JSON-RPC surface could put testnet funds behind a key an
agent had just created. Every published onboarding path ended at registration
and then asked the reader to find funds elsewhere: there was no `lx_*` method,
no CLI command, no SDK call and no MCP tool for a faucet claim, and
`docs/wiki/HostedFaucet.md` documented only the direct route. The row is the
faucet path end to end, bounded per principal per window.

## What changed and why

**Faucet service route.** `POST /v1/faucet/service-claims` admits a trusted
in-cluster caller that presents a Bearer service token, compared in constant
time and only at equal length, together with a principal, a DID and a signer
key. It runs the same preflight as the public claim route, so content type,
forwarded-identity headers, network admission and the idempotency key are
enforced identically, and the body must parse as a JSON object before it is
deserialised so a positional array cannot stand in for a claim. When no service
token is configured the route answers `404 not_found`, so a deployment that has
not opted in exposes nothing new.

**Quota keying.** The caller-supplied principal becomes the identity quota key,
so a principal shares one window with its own direct `POST /v1/faucet/claims`
and cannot claim twice by changing route. Only the network bucket is scoped
`service:<principal>`, so the calling pod's address does not become one global
bucket for every principal behind it.

**Gateway method.** `lx_requestFunds` is served from a new
`platform/hosted/gateway/src/rpc_faucet.rs` with a one-line dispatch hook in
`rpc.rs`. It authenticates with a Bearer identity session rather than a
`LayerX-Key`, precisely so the faucet's identity bucket is shared between the
two routes, and it funds only a key the session already authorises
(`signer_not_bound` otherwise). The claim idempotency key is a domain-separated
length-prefixed SHA-256 over the session principal, the DID and the key, pinned
by test vector, so a retried request is the same claim upstream. Claims are
bounded per minute across the process. Only a claim the faucet confirms as
funded, carrying a hex funding identifier, a transaction identifier, a non-zero
amount and a network, becomes a result; `202` becomes an honest `-32001`
pending refusal, and every other upstream status maps to a typed JSON-RPC code
with a machine-readable `error.data.code`.

**Client surfaces.** `layerx faucet` claims a grant for the active identity and
a named local key. `RpcClient::request_funds` returns a `FaucetGrant` that
cannot be constructed from an unconfirmed claim. `layerx mcp serve` gains a
`faucet.request` tool. `openrpc.json` describes the method with its parameter
patterns and its four refusal codes, taking the document to 17 methods.

**DID well-formedness.** A test caught that `did::alice` was accepted: neither
`layerx_types::ids::Did::new` nor the faucet's `valid_did` rejects an empty
method segment. The three new entry points were tightened to require a
non-empty method and identifier rather than dropping the case.

**Gateway `config()`.** Adding one field to the `Config` literal took the
function to 101 lines and clippy's denied `too_many_lines`. The network
identifier, wire version and protocol network identifier reads moved into
`configured_protocol()` with their checks and error strings unchanged.

**Wiki.** `HostedFaucet.md`, `PublicRpc.md`, `Cli.md` and `RunningAnAgent.md`
document the route, the method, the command, the tool and the quota model, and
the several hundred `path:line` citations that moved because of these
insertions were remapped across thirteen pages.

## Gates

| platform tests | `cargo test --manifest-path platform/Cargo.toml -p layerx-platform-faucet -p layerx-platform-gateway -p layerx-platform-cli` | 0 (153 passed, 0 failed) | `qual-logs/platform-test.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| platform lint | `cargo clippy --manifest-path platform/Cargo.toml -p layerx-platform-faucet -p layerx-platform-gateway -p layerx-platform-cli --all-targets` | 0 | `qual-logs/platform-clippy.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| SDK tests | `cargo test --manifest-path agent/Cargo.toml -p layerx-sdk` | 0 (69 passed, 0 failed) | `qual-logs/agent-sdk-test.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| MCP tests | `cargo test --manifest-path agent/Cargo.toml -p layerx-mcp` | 0 (27 passed, 0 failed) | `qual-logs/agent-mcp-test.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| agent lint | `cargo clippy --manifest-path agent/Cargo.toml -p layerx-sdk -p layerx-mcp --all-targets` | 0 | `qual-logs/agent-clippy.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| ledger | `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| report render | `tools/ci/beta-report.sh` | 0 | `qual-logs/beta-report.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| report check | `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |
| whitespace | `git diff --check` | 0 | `qual-logs/diff-check.log` | `c7aea6b869ee26d3c0565a57b2bcd2dfa5270d24` |

Every platform cargo command ran with `--target-dir` and with `PROGRAMS_CARGO`
pointed at the worktree's own `programs/target`, so the emulator build script
could not recurse into the target directory the outer cargo holds.

## Observations added

`spec/layerx-beta/qualification.kvx`, task `3.6`, ids `4300`-`4307`:

- `4300` the faucet MCP tool was placed in `platform/cli/src/toolset.rs` `SERVED`
  rather than the daemon-routed `TOOL_CATALOGUE`, because the latter needs an
  `Operation` variant and a scope alias in `layerx-agentd` and would otherwise
  advertise a tool that always refuses (note)
- `4301` no hosted manifest configures either end of the service route, so on
  the cluster the method answers `faucet_not_configured` until the token, the
  URL and the NetworkPolicy edge exist (blocker)
- `4302` why the identity bucket is keyed on the principal and only the network
  bucket is scoped to the calling service (note)
- `4303` `PublicRpc.md` claimed 15 methods and omitted `lx_register`, and
  `Cli.md` documents no `layerx register` row (note)
- `4304` the wiki's `path:line` citations are unverified by anything in the
  tree, several gateway ones were already stale before this row, and how the
  moved ones were remapped (suspect)
- `4305` `platform/emulator/build.rs` lets a nested `cargo build` inherit
  `CARGO_TARGET_DIR` and deadlock against its own parent's lock (suspect)
- `4306` `POST /v1/faucet/claims` accepts a positional JSON array in place of
  the documented object, which the new route refuses (note)
- `4307` the `config()` extraction that the `too_many_lines` bound forced (note)


paxlabs-inc:
Rebased mechanically onto main 9f40c1e58: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head ba5f2d92a: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head c7aea6b86 (no file overlap with main since the merge base).

PR #245, opened 2026-09-10
Grow the account registry instead of refusing at 512 accounts

paxlabs-inc:
## Issue

The account registry answered `LXP_ERR_ARENA_EXHAUSTED` on the 513th account. That refusal came from the shape of the storage — a fixed inline `lx_account accounts[512]` — and not from gas, fee or authority, so a node stopped opening accounts for a reason the protocol does not define.

## What changed

`lx_account_registry` is now a handle: a pointer to a slot array, a pointer to a durable sorted index of `{account id, slot}` entries, a committed capacity and a count. Storage is a `PROT_NONE MAP_NORESERVE` reservation taken once and turned readable and writable page by page with `mprotect` as the count grows, doubling from `LX_ACCOUNT_REGISTRY_INITIAL_SLOTS` (512).

A reservation rather than a `realloc` is what makes growth safe: `lx_account` pointers are handed to callers and held across account creation, so a moving allocation would dangle every one of them.

The index is maintained by binary-search insert and remove. Lookups become O(log n) instead of a linear scan, and every consumer that needs accounts in canonical id order — the state-root leaves, the state-diff account list, the snapshot writer — reads that order from the index instead of sorting a copy of the slot array. Snapshot bounds are derived from the index size.

The three capacity refusals in `src/protocol/lxp_module_ctx.c` are gone. What bounds one activity is `LXP_MODULE_MAX_STAGED_ACCOUNTS`. `LX_ACCOUNT_REGISTRY_CAPACITY` remains as the size of the reservation rather than as a rule: reserving past it answers `LXP_ERR_LENGTH_LIMIT` and a failed page commit answers `LXP_ERR_ARENA_EXHAUSTED`, both raised where the memory is acquired.

**The state root does not move.** `leaves_root` already sorted leaves by key, so enumerating in index order builds the same sorted leaf array that enumerating in slot order built. `lxp_test_state_proof --vectors` built from this branch is byte-identical to the same binary built from an unmodified checkout of main (`qual-logs/state-proofs-lane.json` vs `qual-logs/baseline-state-proofs.json`). No chain restart follows.

### Consumers

Consumers are updated to match the new storage, not merely to compile against it:

- Buffers dimensioned from the account capacity are sized from live counts.
- `lx_asset_state_root` moved its leaf and hash arrays off the stack, where the constant would have required a 511 MB frame, and sorts them with a `qsort` whose comparator breaks key ties by original position — reproducing the order the insertion sort produced.
- Registry value copies that would silently have become pointer aliases take `lx_account_registry_copy`, or `lx_account_registry_borrow` where the lifetime is an arena. Without that the replay-divergence and bridge-credit comparisons would have compared a registry with itself.
- The two sites that inserted or removed accounts by writing the slot array directly (`lxp_state_transition_apply`, `lxp_genesis_materialize`) go through the registry helpers so the index cannot go stale.
- Every site that owns a registry releases it.

The shared `program_account_state_v2` vector and `MAX_PROGRAM_VALUE_ACCOUNTS` move with the constant; otherwise the Rust verifier would refuse a valid snapshot of a registry holding more than 512 accounts and reimpose the same bound at the client.

### Tests

`tests/programs/test_accounts.c` gains `registry_growth`, which opens 1543 module-value accounts and checks the count, the committed capacity, the index invariant, every id's lookup, the registry root and sampled account proofs against it.

## Gates

Head `2a544730109cdf9d2d517c5e032e297c96aebeba`, based on `8b62a0f4fa9098a35f1b6c0321e71e60095015c1`.

| command | exit | log |
| --- | --- | --- |
| `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/build.log` |
| `make CC=gcc test` | 0 | `qual-logs/test.log` |
| `make CC=gcc -k` over 13 replay/snapshot/state targets | 2 | `qual-logs/wave11.log` |
| `make CC=gcc -k` over 47 registry-consumer targets | 2 | `qual-logs/registry-consumers.log` |
| `cargo test --manifest-path programs/Cargo.toml -p layerx-programs-registry` | 0 | `qual-logs/programs-registry.log` |
| `make beta-ledger-check` | 0 | `qual-logs/ledger-check.log` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report.log` |
| `git diff --check` | 0 | `qual-logs/diff-check.log` |

Every gate above ran against `2a544730109cdf9d2d517c5e032e297c96aebeba`.

Both non-zero exits are `make -k` aggregates over failures that are not this change:

- **13-target batch**: `test-state-proof` only. `contracts/config/native-state-proofs.json` is stale in a module-registration leaf. This branch's `--vectors` output equals unmodified main's. The target is not part of `make test`, which passes.
- **47-target batch**: `test-genesis-bootstrap`, `test-asset-withdraw-activity`, `test-finality-evidence`, `test-bridge-credit`. All four fail identically on an unmodified checkout of main built in a separate worktree (`qual-logs/baseline-consumers.log`, same four targets, same messages). Every other target in the batch passes, including `test-replica`, `test-replay`, `test-snapshot`, `test-state-root`, `test-state-diff`, `test-state-commitment-transition`, the six perps targets, and the escrow, stream and gateway targets.

## Observations added

`spec/layerx-beta/qualification.kvx`, task 6.11, blocks 2300 through 2308: the storage and index design and the unchanged state root (2300), the refusal surface that replaced the capacity bound (2301), the out-of-path consumer repairs (2302), the index-maintenance sites (2303), the shared vector and Rust constant (2304), the stale state-proof vector (2305), the genesis-bootstrap failure (2306), the `lx_asset_state_root` stack frame (2307), and the four pre-existing consumer failures with their baseline evidence (2308).

## Build and test after rebase

The gates in the table above ran at `2a544730109cdf9d2d517c5e032e297c96aebeba`, based on `8b62a0f4fa9098a35f1b6c0321e71e60095015c1`, before the branch was rebased. They were not re-run after it.

A later local run of the same gate script at `03c05527ebc869e373d71adcce05f95eeb316e21` (based on `dfd9618feea31a595cacb4eeb30ed92a6e936d8d`; the same code tree as this branch minus the rebase and the second commit) is what `qual-logs/` on the lane workstation now holds: `make CC=gcc build/bin/layerxd build/bin/layerx-genesis-build` 0, `make CC=gcc test` 0, the 13-target batch 2 (`test-state-proof` only, as above), `cargo test -p layerx-programs-registry` 0, `make beta-ledger-check` 0, `tools/ci/beta-report.sh --check` 0, `git diff --check` 0. The 47-target consumer batch in that run was terminated by disk exhaustion (`ld: final link failed: No space left on device`, exit 137) and did not complete, so its exit-2 result with the four pre-existing failures stands only at the revision in the table.

The branch was then rebased onto `b3e57a437ddadbb7be908357920b9be4a6646acd` and one commit was added (`a7eaa4537a3f660a9b80d625c969003e27bf8034`). Nothing was built, compiled or tested after that: no gate above was re-run on the rebased head, and the added commit has not been compiled or exercised by any gate.

Rebase resolution: `spec/layerx-beta/qualification.kvx` is main's exact bytes followed by blocks 2300 through 2308, separated by the blank line the file uses between blocks (the textual auto-merge had dropped only that separator); `spec/layerx-beta/report.md` is main's version re-rendered by `tools/ci/beta-report.sh`; `cmd/layerxd/lxp_daemon_lni.c` auto-merged with main's version-refusal change and this branch's `send_did_accounts` buffer change both intact.

### Added commit: emulator snapshot import

`platform_emulator_snapshot_import` in `platform/emulator/core/emulator_core.c` copied the wrapper snapshot's account bytes straight into `accounts.accounts` and set the count. Under the new storage that writes through a null slot pointer on an emulator that has never opened an account, and otherwise leaves the id index empty so every later lookup and state root refuses the registry as non-canonical. It now reserves `header.account_count` slots, copies, sets the count and rebuilds the index, the sequence `lxp_snapshot_load` uses, and `platform_emulator_destroy` releases the registry mapping alongside the state store. The emulator core is compiled by the platform workspace's `build.rs`, not by `make test`, which is why the gates above did not reach it.

### Static checks run after the rebase

| command | exit |
| --- | --- |
| `python3 kvx_verify.py <origin/main ledger> spec/layerx-beta/qualification.kvx` | 0 (`missing=[] differ=[]`, nine extra blocks) |
| `git diff origin/main HEAD --numstat -- spec/layerx-beta/qualification.kvx` | `72 0` |
| `tools/ci/beta-report.sh` | 0 |
| `tools/ci/beta-report.sh --check` | 0 |
| `git diff --check origin/main HEAD` | 0 |
| `rustfmt --edition 2021 --check programs/crates/layerx-programs-registry/src/account_state.rs programs/crates/layerx-programs-registry/tests/winddown.rs` | 0 |

### Generated artifacts pending

None. `contracts/config/native-state-proofs.json` is stale independently of this change (observation 6.11.2305) and is left as it is.

### Follow-ups

- `platform/emulator/src/main.rs` keeps `MAX_EMULATOR_ACCOUNTS = 512` as the emulator's own prefund and recovery-snapshot bound, next to its 4096-entry move and program bounds. The emulator's C core no longer limits accounts, so that constant is now the only place the emulator refuses a 513th prefund (`Err(-5)`).
- `lx_account_list_did` enumerates a DID's accounts in account-id order from the index rather than in slot (creation) order, so the LNI account-read listing for a DID is id-sorted.


paxlabs-inc:
Rebased mechanically onto main 123d797bb: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head e2e874bb7: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head a7eaa4537 (no file overlap with main since the merge base).

PR #247, opened 2026-09-10
Check the hosted topology by named edge and run the agentd checks

paxlabs-inc:
## Issue

`make platform-hosted-topology-check` was red on an unmodified checkout, and the two hosted agentd surfaces had no check target.

`platform/hosted/tests/topology-regressions.py` pinned the resolved topology as two integers — forty resolved edges without the Human manifest and forty two with it — while the tree resolves forty one and forty three. `platform/hosted/agentd/manifest_check.py` had no caller anywhere in the tree, and `platform/hosted/agentd/probe.sh` was reached only from `platform/hosted/tests/beta-cluster.sh:1720`, which needs a cluster, so neither ran in any check.

## What changed and why

**The topology regression tests edges, not numbers.** Before re-pinning anything, the checker and all nine manifests were reconstructed at the revision that wrote those integers and the resolved edge set was diffed against the tree. Exactly one resolved edge was added:

```
Deployment layerx-testnet/layerx-gateway -> layerx-pending-core.layerx-testnet.svc.cluster.local:9443 [env LAYERX_GATEWAY_PUBLIC_CORE_URL]
```

It is the gateway's public read source: declared at `platform/hosted/gateway/deployment.yaml:78`, documented at `platform/hosted/gateway/README.md:13` and `docs/wiki/HostedGateway.md:222` and `:245`, admitted by the `layerx-gateway` egress rule for `layerx-plane: trusted-boundary` on 9443 and by `layerx-node-ingress` at the node pod, reaching a declared Service whose `targetPort: core-tls` is a container port of that pod. Observation `6.2.919` records the same edge being admitted by the checker when it was added. One `note` edge also moved (`LAYERX_HUMAN_WEB_ORIGIN`, from the Human Deployment to the node StatefulSet); it resolves off cluster and is not counted.

Both integers are replaced by named tuples:

- `RESOLVED_EDGES` — the forty one edges the eight manifests resolve
- `HUMAN_EDGES` — the two the Human manifest resolves, which the baseline must leave failing
- `SEPARATELY_OPERATED_EDGES` — the status publisher edge

`named()` compares the checker's labelled rows against a tuple and reports duplicates, edges the manifests resolve that nobody named, and named edges that disappeared, before asserting the length. Every count the regression asserts is now `len()` of a named tuple, and the two Human failures are asserted by name instead of by the number two. An edge label carries the caller workload, the host, the port and the configuration key that declares it, so an addition is now a named diff a reviewer can approve.

No assertion was relaxed. The nine mutation refusals, the parser parity check, the namespace precedence case and the node literal block case are unchanged; the mutation bound `len(result) > 2` now reads `len(result) > len(HUMAN_EDGES)`, the same bound expressed through the named set.

**The agentd surfaces run.** `platform/Makefile.inc` declares `platform-hosted-agentd-check`, which parses `probe.sh` with `sh -n` and runs `manifest_check.py`, and `platform-hosted-topology-check` depends on it. `manifest_check.py` reads the node manifest and `platform/hosted/human/material.py` from the tree and needs nothing deployed, so it runs in full — proving the `layerx-agentd` Service, the mutually authenticated boundary, the loopback port disjointness of the node pod and the ingress closure, with thirteen mutations refused. `probe.sh` asserts properties of a deployed Service, so the target parses it and leaves the live assertions to the bring-up that already calls it.

## Gates

| --- | --- | --- | --- |
| `make platform-hosted-topology-check` | 0 | `qual-logs/topology-check.log` | 69997caf4 |
| `python3 platform/hosted/agentd/manifest_check.py` | 0 | `qual-logs/agentd-manifest-check.log` | 69997caf4 |
| `sh -n platform/hosted/agentd/probe.sh` | 0 | `qual-logs/agentd-probe-parse.log` | 69997caf4 |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | 69997caf4 |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | 69997caf4 |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | 69997caf4 |
| negative: extra unnamed manifest edge | 1 (refusal expected) | `qual-logs/negative-unnamed-edge.log` | 69997caf4 |

The negative case injected one extra env URL into the gateway manifest and confirmed the regression refuses it by name rather than by count:

```
AssertionError: ('baseline resolved edges', "unnamed: ['Deployment layerx-testnet/layerx-gateway -> layerx-identity.layerx-testnet.svc.cluster.local:9443 [env LAYERX_GATEWAY_UNNAMED_PROBE_URL]']", 'named but absent: []')
```

The manifest was restored and the tree left clean.

## Observations added

- `observation.3.6.7500` — the named edge sets replacing the two pinned integers, the single added edge and the evidence that it is an intended documented connection.
- `observation.3.7.7501` — the agentd check target and why the probe is parsed rather than executed.

`spec/layerx-beta/report.md` was re-rendered by `tools/ci/beta-report.sh` in the same commit; it was not edited by hand.


paxlabs-inc:
Rebased mechanically onto main 4f4126847: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head d6985a3c7: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 69997caf4 (no file overlap with main since the merge base).

PR #249, opened 2026-09-10
Add WebSocket unsubscribe and a per-topic resume cursor

paxlabs-inc:
## Issue

The public WebSocket surface had no way to cancel one subscription and no reconnect
contract. `platform/hosted/gateway/src/ws.rs` identified a subscription by its
position in a `Vec`, so removing one would have renumbered the survivors, and a
notification carried only `params.subscription` and `params.result`, so a client
that lost its socket could not say where its stream stopped. `openrpc.json` said as
much: notifications were live, not a durable replay.

## What changed and why

**Gateway (`platform/hosted/gateway/src/ws.rs`)**

- `lx_unsubscribe` takes the identifier issued on that connection and answers
  `{"result": true}`. Identifiers now come from a per-connection counter
  (`Subscriptions::next`) instead of a position, so the remaining subscriptions keep
  their identifiers and cursors and a released identifier is never reissued. An
  identifier that is not active on the connection is refused with `-32602`, and
  `POST /rpc` refuses `lx_unsubscribe` with `-32004` exactly like `lx_subscribe`.
  The release itself is one function, `cancel()`, called by the frame handler and by
  its test, so the test exercises the code the socket runs rather than a copy of it.
- Every notification carries `params.cursor`: the receipt-feed position at which it
  was delivered, as a canonical decimal string. The hub publishes `(position, event)`
  pairs, so a receipt is stamped with its own `global_sequence` and an idle wake with
  the position the feed has consumed.
- `lx_subscribe` accepts that cursor as a trailing positional argument on all three
  topics (`["receipts", "41"]`, `["checkpoints", "41"]`,
  `["account", account_id, "41"]`). On resume the gateway reads
  `chain_head_sequence`, refuses a cursor above it with `-32602` and a gap wider than
  the 16-position window with `-32005` **without creating the subscription**, then
  replays through the same notification path it uses live: every receipt event after
  the cursor for `receipts`, and the current snapshot once for `checkpoints` and
  `account`. Replayed frames are written after the acknowledgement and before any
  live frame, and `advance()` drops a live receipt at or below the cursor so the
  overlap between the replay and the running feed is never delivered twice.
- `receipt_event()` is factored out of the feed loop so the live feed and the replay
  read the core receipt-event endpoint through one definition of that contract.

**Schema and docs** — `openrpc.json` gains the `lx_unsubscribe` entry and the
`cursor` parameter on `lx_subscribe`, and its description now states the resume
contract, the window and the refusals. `docs/wiki/PublicRpc.md` and
`docs/wiki/PublicAPI.md` carry the same rows and the notification shape;
`PublicAPI.md` also gains the `lx_register` and `lx_requestFunds` rows it was
missing, so both tables and the schema describe the same 18 method names.

**SDKs** — both notification decoders moved with the third `params` key and refuse a
notification whose cursor is absent or non-canonical rather than defaulting it.
`RpcClient::subscribe_from` / `JsonRpcClient.subscribeFrom` resume from a cursor;
`RpcSubscription::unsubscribe` drains at most 16 in-flight notifications, advancing
the cursor past them, before it reads the acknowledgement, and the TypeScript
generator sends `lx_unsubscribe` on teardown over a still-open socket and waits up to
five seconds for that acknowledgement.

One hunk outside the listed owned paths was unavoidable and is minimal:
`platform/hosted/gateway/src/rpc.rs` dispatches `lx_unsubscribe` to the `-32004`
WebSocket refusal, and its `schema_lists_every_public_method` gains the new name.
That assertion previously carried the method count as a literal beside the list of
names it checks, so two changes collided on the same line here — `lx_unsubscribe`,
and `lx_requestFunds` arriving at a rebase. It now declares the published set once
and compares the schema's length against that set, so the count is derived from the
declaration instead of written down twice, and a later method addition settles one
line. The check is not weakened: an undeclared method in the schema still fails it,
and a declared method missing from the schema still names itself in the failure.

Nothing in `platform/hosted/gateway/tests/local/lifecycle.rs` was touched: it pins
the first three acknowledgements as `"1"`, `"2"`, `"3"` and the notifications'
subscription as `"1"` and `"3"`, which the counter reproduces exactly, and it does
not bound the size of `params`.

## Gates

Every gate below ran against `089fda0b0`, the head of this branch.

| --- | --- | ---: | --- |
| format | `rustfmt --edition 2021 --check` on the four touched Rust files | 0 | `qual-logs/rustfmt-check.log` |
| gateway tests | `CARGO_TARGET_DIR=/root/lx-shared-target/platform cargo test --manifest-path platform/Cargo.toml -p layerx-platform-gateway` | 0 | `qual-logs/gateway-tests.log` |
| sdk tests | `CARGO_TARGET_DIR=/root/lx-shared-target/agent cargo test --manifest-path agent/Cargo.toml -p layerx-sdk` | 0 | `qual-logs/sdk-tests.log` |
| workspace install | `npm ci --workspaces --include-workspace-root --ignore-scripts --no-audit --no-fund` | 0 | `qual-logs/npm-ci.log` |
| typescript sdk | `npm test` in `agent/sdk/typescript` | 0 | `qual-logs/sdk-typescript-npm-test.log` |
| ledger check | `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` |
| report render | `tools/ci/beta-report.sh` then `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report.log`, `qual-logs/beta-report-check.log` |
| whitespace | `git diff --check` | 0 | — |

The `layerx-sdk` run above is the second one at this head: the first died with
`No space left on device` while linking `layerx-programs-runtime`, and a build that
hit ENOSPC is not recorded as a gate. It was re-run in full once the volume had room
and is the run reported here.

Not run: `local_gateway_websocket_receipt_wake` in the standalone harness at
`platform/hosted/gateway/tests/local`. It needs the native daemon and four platform
binaries built first and its own serialized target directory, and that directory and
the shared platform target were held by other qualification runs throughout this
work. Recorded as observation `6.2.4703` rather than claimed.

Each rebase onto `main` reproduced the known `qualification.kvx` folding hazard: the
final block on `main` lost its `severity = "note"` line and the blank separator to
the first appended block, with no conflict marker and zero deleted lines in
`git diff`. The ledger was rebuilt each time as `main`'s exact content followed by
this branch's four blocks, and `kvx_verify.py` reports `missing=[] differ=[]`
against `origin/main` at this head.

## Observations added

`spec/layerx-beta/qualification.kvx`, task `6.2`:

- `6.2.4700` — the cancellation and resume contract, the cursor's meaning per topic,
  and why the window is 16.
- `6.2.4701` — the OpenRPC method-set assertion, why the count is now derived from
  the declared set, and the harness assertions the identifier counter preserves.
- `6.2.4702` — the wiki method table that omitted `lx_register` and
  `lx_requestFunds` and the count sentences that disagreed with the schema, both
  completed here.
- `6.2.4703` — the real-socket gate that did not run on this revision.

`spec/layerx-beta/report.md` is re-rendered from the ledger by
`tools/ci/beta-report.sh` in the same commit.


paxlabs-inc:
Rebased mechanically onto main c2b212c1d: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head ca553f4c8: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 089fda0b0 (no file overlap with main since the merge base).

PR #250, opened 2026-09-11
Give the nested programs build its own cargo target directory

paxlabs-inc:
## The issue

`platform/emulator/build.rs` runs `make build` in the repository root. `make build`
reaches `programs-build` through the order-only `| programs-build` prerequisite that
`build: layerx-guarantor` pulls in, and `programs-build` ran `cd programs && $(PROGRAMS_CARGO) build`
with `PROGRAMS_CARGO ?= cargo`. A build script inherits its caller's environment, so when the
outer cargo had `CARGO_TARGET_DIR` set it also held that directory's `debug/.cargo-lock`, and the
nested cargo blocked on the lock its own ancestor was holding. Neither could finish, and every
other build sharing that target directory queued behind the pair.

## What changed

`Makefile` (the `PROGRAMS_CARGO` block and the two programs-workspace build recipes only):

- `PROGRAMS_TARGET_DIR := $(CURDIR)/programs/target` names the directory the programs workspace
  builds into - the same file `PROGRAMS_RUNTIME_LIB` is read from.
- `PROGRAMS_WORKSPACE_CARGO = env CARGO_TARGET_DIR='$(PROGRAMS_TARGET_DIR)' $(PROGRAMS_CARGO)`
  is what `programs-build` and the `$(PROGRAMS_RUNTIME_LIB)` rule now run, so the nested build
  always owns its own target directory and can never contend with its caller's.
- `PROGRAMS_CARGO` keeps its `?=` default and its meaning - the cargo program to run - so the
  pinned-toolchain overrides in `.github/workflows/programs-conformance.yml` still select the
  toolchain, and an explicit `PROGRAMS_CARGO='env CARGO_TARGET_DIR=... cargo'` still wins.
- `PROGRAMS_RUNTIME_LIB` is unchanged and still spelled relative, so
  `make programs/target/debug/liblayerx_programs_sandbox.a` - which
  `agent/crates/layerx-crypto/tests/generate-native-payments.sh` invokes by that exact name -
  still matches the rule.

`platform/emulator/build.rs` additionally removes `CARGO_TARGET_DIR` and `CARGO_BUILD_TARGET_DIR`
from the environment it hands to `make`, so the nested build cannot inherit the caller's target
directory even through a `PROGRAMS_CARGO` override that does not set one.

The programs dependency is intact, the runtime library is real, and the emulator's link line is
untouched.

The target-directory pin is deliberately a separate variable rather than folded into
`PROGRAMS_CARGO`: folding it in would also force the three `programs-porting-v2-references`
reference crates - each its own `[workspace]` - out of the `porting/*/reference-v2/target`
directories the lint and activity steps read back.

## Gates

| Gate | Exit | Log | Revision |
| --- | --- | --- | --- |
| `cd platform && cargo fmt -p layerx-platform-emulator --check` | 0 | `qual-logs/fmt.log` | `033c9b47d` |
| `make -n build` | 0 | `qual-logs/make-n-build.log` | `033c9b47d` |
| `make CC=gcc build/bin/layerx-guarantor` | 0 | `qual-logs/make-guarantor.log` | `033c9b47d` |
| `make CC=gcc build` | 0 | `qual-logs/make-build-full.log` | `033c9b47d` |
| `cargo check -p layerx-platform-emulator` from `platform/` with `CARGO_TARGET_DIR` set to a scratch directory in the worktree | 0 | `qual-logs/emulator-check.log, qual-logs/emulator-check-pstree.log` | `033c9b47d` |
| deadlock probe: a held probe `debug/.cargo-lock` plus `CARGO_TARGET_DIR` pointing at it, `make build` on the parent revision | 137 (still blocked when its budget expired - the reproduction) | `qual-logs/lock-probe.log` | `033c9b47d` |
| same probe, same environment, `make build` on this revision | 0 | `qual-logs/lock-probe.log` | `033c9b47d` |
| `make -n programs-build` with `CARGO_TARGET_DIR` exported, parent revision vs this one | 0 | `qual-logs/programs-target-dir-before-after.log` | `033c9b47d` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `033c9b47d` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `033c9b47d` |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | `033c9b47d` |

## Observations added

- `[observation.4.3.8900]` (note) - the deadlock, the repair, and the evidence for both.
- `[observation.4.3.8901]` (suspect) - `programs-interpreter-conformance`,
  `programs-interpreter-bench` and `programs-porting-v2-references` still honour an inherited
  `CARGO_TARGET_DIR` while reading their artifacts back from a fixed path; recorded, not repaired,
  because the fix belongs to those recipes and needs a per-invocation `--target-dir`.


paxlabs-inc:
Rebased mechanically onto main 119594e88: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 8f10a4b46: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 033c9b47d (no file overlap with main since the merge base).

PR #253, opened 2026-09-11
Persist the identity tenant and scope principals and sessions to it

paxlabs-inc:
## Issue

`platform/hosted/identity` required, validated and echoed a tenant on `POST /v1/principals` but discarded it before storage. `Principal` held only the subject, and the principal map was keyed by the subject alone, so provisioning the same subject under a second tenant silently replaced the first tenant's signer key set, account and audiences. Sessions carried no tenant at all, and introspection resolved a session's principal by subject, so nothing in the service was scoped to a tenant.

## What changed and why

**Tenant is persisted and part of the key.** `Principal` and `StoredSession` now carry a `tenant`. The store keeps principals in a tenant → subject map, so a principal is addressed as `(tenant, sub)`. The snapshot, every journal record and the replay path carry the tenant, and `Store::open` refuses a snapshot whose principal is filed under a tenant other than its own, a snapshot or journal that binds one subject to two tenants, a principal with no tenant, and a session naming a principal outside its own tenant.

**Cross-tenant replacement is refused, not silently applied.** The store holds a subject → tenant binding and refuses any principal whose subject is already bound elsewhere. `POST /v1/principals` answers that with `409 subject_tenant_conflict` instead of overwriting the first tenant's record; the existing principal keeps its signer keys, account and audiences. Refusing rather than merely keying is the stricter choice and is sound because the gateway mints subjects as a tenant-domain-separated hash of the signer key (`platform/hosted/gateway/src/rpc_register.rs`), so a subject already identifies exactly one tenant.

**Session create and lookup are tenant-scoped.** `POST /v1/sessions` accepts an optional `tenant`; when present it is validated and the request is refused with `404 principal_not_found` unless the subject is bound to exactly that tenant, and when absent the tenant is resolved from the store's binding. The minted session records that tenant and the response names it. The per-principal live-session bound counts sessions of `(tenant, sub)`. Introspection resolves the principal as `(session.tenant, session.principal)`, so a session can never be answered with another tenant's principal.

The four introspection response shapes are deliberately unchanged, so `gateway`, `webhooks`, `dashboard`, `faucet`, `testnet` and the ramp consumer are unaffected. Observation 6.5.5104 records the constraint that makes identifying a session by subject alone safe.

**Tests.** New store tests cover tenant/subject keying, session scoping, and six ways cross-tenant state on disk is refused. A new service test provisions two tenants over TLS against the real binary and proves that a cross-tenant claim is a 409, that a session is refused for a subject of another tenant, that a bad tenant name is a 400, that the resolved and explicitly scoped paths agree, that each session introspects only to its own tenant's signer keys and account, and that all of it survives a restart through the compacted snapshot.

## Gates

|---|---|---|---|---|
| Crate tests (21 unit + 10 service) | `CARGO_TARGET_DIR=/root/lx-shared-target/platform cargo test --manifest-path platform/Cargo.toml -p layerx-platform-identity` | 0 | `qual-logs/cargo-test-identity.log` | `7d00ae54b07f9a7ef8e4a532409c913e43788492` |
| Format | `cargo fmt --manifest-path platform/Cargo.toml -p layerx-platform-identity -- --check` | 0 | `qual-logs/cargo-fmt-check.log` | `7d00ae54b07f9a7ef8e4a532409c913e43788492` |
| Lints | `cargo clippy --manifest-path platform/Cargo.toml -p layerx-platform-identity --all-targets` | 0 | `qual-logs/cargo-clippy-identity.log` | `7d00ae54b07f9a7ef8e4a532409c913e43788492` |
| Ledger | `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `7d00ae54b07f9a7ef8e4a532409c913e43788492` |
| Report render | `tools/ci/beta-report.sh` | 0 | `qual-logs/beta-report.log` | `7d00ae54b07f9a7ef8e4a532409c913e43788492` |
| Report check | `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `7d00ae54b07f9a7ef8e4a532409c913e43788492` |

Every gate above was run on this revision, which is based on `b01a0d6df`.

## Observations added

- `6.5.5100` — the defect and the shape of the fix; why a global subject→tenant binding is the stricter reading of "cross-tenant replacement refused".
- `6.5.5101` — the on-disk snapshot shape changed and there is no automatic upgrade; an existing identity state directory fails to load and the service exits 2 rather than starting with a partial view.
- `6.5.5102` — why `tenant` on `POST /v1/sessions` is optional: two callers outside this lane's owned paths mint sessions with a subject only.
- `6.5.5103` — an existing test asserted that one subject could be created under two tenants; what replaced it and why the replacement proves at least as much.
- `6.5.5104` — introspection shapes still identify a session by subject alone; the constraint that keeps that safe.


paxlabs-inc:
Rebased mechanically onto main fffdef4ec: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 57b933ca1: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 7d00ae54b (no file overlap with main since the merge base).

PR #254, opened 2026-09-11
Charge the resolved grant on every delegated debit leg

paxlabs-inc:
## Issue

The transfer transition never consulted an authority allowance. `lxp_authority_charge_allowance` existed in `src/state/lxp_authority.c` but nothing in the ledger called it, so a debit that declared `LXP_AUTH_DELEGATED_CAPABILITY` moved balances with no grant consulted at all, and a grant's per-activity, lifetime and period caps were dead state.

## What changed and why

`lxp_transfer_context` now carries an `lxp_transfer_allowance` holding a pointer to the caller's live `lxp_authority_scope`, the grant kind and the grantor. The caller keeps ownership of the scope, so a caller that resolves a grant, executes, and writes the same scope back to the grant record needs no new state.

Two new authority entries in `src/state/lxp_authority.c` do the binding:

- `lxp_authority_check_debit` reads the scope without touching it. It refuses a module outside the grant `module_mask`, an asset the grant does not meter, and a draw past any cap. The period roll runs on a copy, so checking can never advance the window.
- `lxp_authority_charge_debit` performs the same binding, then calls `lxp_authority_charge_allowance`, which remains atomic on a copy: an exhausted draw leaves the scope byte-identical.

A kind that carries no allowance (`OWNER`, `SESSION_KEY`, `ESCROW`, `PROTOCOL_MODULE`) must present a scope with no caps and no recorded spend, so a metered grant cannot be spent through an unmetered kind.

In the ledger, check and charge are deliberately separate. The check runs inside `lxp_precondition_check`, where every caller already reads it as a pure predicate with a `const` context. The charge is a separate internal entry, `lxp_allowance_charge_leg`, declared in the private `src/ledger/lxp_ledger_internal.h` (alongside `lxp_balance_restore_snapshot` and `lxp_sequence_terminal_check`, so a module object cannot reach it). It is called once per leg immediately before `lxp_balance_apply_leg`, from both `lxp_apply_transfer` and the transfer-set loop in `src/ledger/lxp_journal.c`.

Rollback safety: the ledger journal records the pre-charge scope on the first charge of a set, and `lxp_balance_restore_snapshot` restores it before it restores any account, so a set that fails at any leg leaves the recorded spend exactly where the balances go back to. The single-leg path snapshots and restores the scope around its own balance apply.

Refusals are typed by the authority domain and terminal-rejection applicable:

| Condition | Refusal |
| --- | --- |
| Delegated debit presents no grant | `LXP_ERR_AUTH_ALLOWANCE` |
| Module outside the grant mask, or presented grant kind does not match the debit, or a metered scope drawn as an unmetered kind | `LXP_ERR_AUTH_SCOPE` |
| Asset the grant does not meter | `LXP_ERR_ASSET_MISMATCH` |
| Debited account is not the grantor | `LXP_ERR_UNAUTHORIZED_DEBIT` |
| Per-activity, lifetime or period cap | `LXP_ERR_GRANT_EXHAUSTED` |

One explicit carve-out: the Budget module charges its own `lx_budget_record` before it emits the set (`src/modules/budget/lx_budget_spend.c:119`, `lx_budget_close.c:55`), so a `LXP_AUTH_BUDGET_ALLOWANCE` debit whose `origin_module_id` is `LXP_MODULE_BUDGET` is admitted without a grant scope. Every other origin, and every delegated debit, must present one. Two independent meters for one budget draw would disagree; the carve-out is commented and is a single line to delete if Budget is later moved onto a persisted grant.

Tests exercise real code paths and real types throughout:

- `tests/state/lxp_test_allowance.c` covers the two authority entries directly (null arguments, unknown kinds, module mask including index 64, asset binding, per-activity, lifetime and period caps, the check never mutating, unmetered kinds refusing any cap or recorded spend), and resolves a capability grant out of persisted governance KV state through `lxp_authority_resolve_activity`, then charges the resolved scope to exhaustion to prove the bound is real rather than unlimited.
- `tests/ledger/test_apply_transfer.c` drives the single-leg path through every refusal and both caps, asserting balances, sequence and the scope are byte-unchanged on each refusal, and covers both sides of the Budget carve-out.
- `tests/ledger/test_transfer_set.c` drives a two-leg set: charges accumulating across legs, a cap breach on the second leg rolling back the first leg's charge, the injected-failure path restoring the spend along with the balances, and a delegated set with no grant refused before anything moves.

## Gates

| --- | --- | --- | --- |
| `make CC=gcc test` | 0 | `qual-logs/test.log` | 3cbc3c9d8 |
| `make CC=gcc beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | 3cbc3c9d8 |
| `tools/ci/beta-report.sh` | 0 | `qual-logs/beta-report.log` | 3cbc3c9d8 |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | 3cbc3c9d8 |
| `git diff --check origin/main HEAD` | 0 | `qual-logs/diff-check.log` | 3cbc3c9d8 |
| `build/tests/lxp_test_allowance`, `build/tests/test_apply_transfer`, `build/tests/test_transfer_set` | 0 | `qual-logs/focused-allowance.log` | 3cbc3c9d8 |

## Observations added

`spec/layerx-beta/qualification.kvx`, task 2.2:

| Record | Severity | Subject |
| --- | --- | --- |
| `observation.2.2.2200` | note | The shape of the change: check separated from charge, where each runs, and the full typed refusal set. |
| `observation.2.2.2201` | blocker | The daemon resolves a grant into a stack local at three sites in `cmd/layerxd/lxp_daemon_process.c` and discards it; no caller yet presents the scope through `lxp_transfer_context.allowance` or persists a charged scope back to the tag-5 grant record, so a delegated draw the ledger charges is invisible to the next activity. Outside this row's owned paths. |
| `observation.2.2.2202` | suspect | Two outer unwind paths (`restore_transfer_snapshots` in `src/protocol/lxp_module_ctx.c`, the gateway journal in `src/network/lxp_gateway_send.c`) restore balances only and would keep a committed charge. Both files outside this row. |
| `observation.2.2.2203` | note | Why the Budget carve-out exists and what deletes it. |
| `observation.2.2.2204` | note | `src/ledger/lxp_send.c` and `lxp_receive.c` take the debit authorization kind straight from the decoded activity; the declared-kind bypass is now refused, but making a delegated send work needs those paths to resolve and bind the grant. |
| `observation.2.2.2205` | note | The three files touched outside the row's listed owned paths and why each field cannot live elsewhere. |

`spec/layerx-beta/report.md` is regenerated by `tools/ci/beta-report.sh`, not hand-edited.


paxlabs-inc:
Rebased mechanically onto main 755e9b6cf: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head b7594b059: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 3cbc3c9d8 (no file overlap with main since the merge base).

PR #257, opened 2026-09-11
Wire the hosted gateway to identity provisioning and to the faucet

paxlabs-inc:
## Issue

The hosted gateway Deployment was not wired to either service the two public
onboarding methods call, so both answered "not configured" in the shipped
cluster.

`lx_register` creates its principal through `POST /v1/principals`, which the
identity service admits only for the `provisioning` service token
(`platform/hosted/identity/src/main.rs:874-879`). The gateway pod held only its
own gateway-scoped token, so the method answered `-32001` with `data.code`
`registration_not_configured` (observation `6.5.4200`, blocker).

`lx_requestFunds` proxies to `POST /v1/faucet/service-claims`, admitted only by
a shared service token that no manifest provisioned, and no NetworkPolicy
admitted the gateway at the faucet pod. The gateway answered `-32001`
`faucet_not_configured` and the faucet answered `404 not_found` for that route
(observation `3.6.4301`, blocker).

## What changed and why

**Identity provisioning token into the gateway pod.** The Deployment sets
`LAYERX_GATEWAY_IDENTITY_PROVISIONING_TOKEN_FILE` to
`/run/layerx/identity-provisioning/token`
(`platform/hosted/gateway/deployment.yaml:85`) and mounts that directory read
only from the existing `layerx-identity-service-tokens` Secret, projecting
exactly one key with `items: [{key: provisioning, path: token}]` (lines 114 and
127). The projection is deliberate: that Secret also carries the gateway,
webhooks, dashboard, faucet, testnet and ramp tokens
(`platform/hosted/tests/beta-cluster.sh:556-564`), and a whole-Secret mount
would hand every one of them to the gateway pod.

**Gateway-to-faucet service claim.** One Secret, `layerx-gateway-faucet-client`,
carries the shared token, mirroring `layerx-gateway-component-client`, which the
gateway, the pending core and the agent boundary already share the same way.
Bring-up generates it with the existing `write_token` helper
(`platform/hosted/tests/beta-cluster.sh:521`) and applies it beside the other
gateway client tokens (line 790). The gateway mounts it at `/run/layerx/faucet`
and sets `LAYERX_GATEWAY_FAUCET_URL` and
`LAYERX_GATEWAY_FAUCET_SERVICE_TOKEN_FILE`
(`platform/hosted/gateway/deployment.yaml:86-87`, `:115`, `:128`); the faucet
mounts the same Secret at `/run/layerx/gateway-faucet` and sets
`LAYERX_FAUCET_SERVICE_CLAIM_TOKEN_FILE`
(`platform/hosted/testnet/deployment.yaml:147`, `:168`, `:175`). No token value
appears in any manifest; both ends read the mounted file, and both trim the
trailing newline a Secret projection may add
(`platform/hosted/gateway/src/main.rs:286-297`;
`platform/hosted/faucet/src/main.rs:218-223`).

The URL is `https://layerx-faucet-public.layerx-testnet.svc.cluster.local:443`,
the one spelling the faucet server certificate is issued for
(`platform/hosted/tests/beta-cluster.sh:395-396`), which the gateway verifies
against `LAYERX_GATEWAY_OUTBOUND_CA_DER`; any other spelling would fail the
handshake. Port 443 is the port `Service layerx-faucet-public` declares, and its
`targetPort: https` is container port 9443 of the faucet pod.

**Policy on both ends.** A gateway egress rule to `app=layerx-faucet` on TCP
9443 (`platform/hosted/gateway/deployment.yaml:191-192`) and an explicit
`layerx-faucet-public-ingress` rule from `app=layerx-gateway` on TCP 9443
(`platform/hosted/testnet/deployment.yaml:310-311`). The ingress rule is
explicit rather than leaning on the `0.0.0.0/0` rule beside it, so the gateway
keeps reaching the faucet if that public rule is ever narrowed; negative case 1
below proves the new rule is the one that admits the edge.

**The new edge is named, not counted.** `RESOLVED_EDGES` gains
`Deployment layerx-testnet/layerx-gateway -> layerx-faucet-public.layerx-testnet.svc.cluster.local:443 [env LAYERX_GATEWAY_FAUCET_URL]`
(`platform/hosted/tests/topology-regressions.py:31`), as the named-edge
regression from PR 247 requires. No count literal was touched: every count the
regression asserts is `len()` of a named tuple, and the checker now resolves 44
named edges instead of 43.

The faucet Deployment, its Service and its NetworkPolicies live in
`platform/hosted/testnet/deployment.yaml`, not in
`platform/hosted/faucet/`, which holds only `Cargo.toml`, `Dockerfile` and
`src/`; that is the file lane U3's evidence also names.

## Gates

All gates ran on `417f84a8d9f18f3533ebe350dec0a358739a55fb`.

| --- | --- | --- | --- |
| hosted topology | `make platform-hosted-topology-check` | 0 | `qual-logs/topology-check.log` |
| topology regressions | `python3 platform/hosted/tests/topology-regressions.py` | 0 | `qual-logs/topology-regressions.log` |
| bring-up parse | `bash -n platform/hosted/tests/beta-cluster.sh` | 0 | `qual-logs/beta-cluster-parse.log` |
| ledger | `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` |
| report | `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` |
| whitespace | `git diff --check` | 0 | `qual-logs/git-diff-check.log` |
| manifest assertions | `python3 qual-logs/manifest-assert.py` | 0 | `qual-logs/manifest-assert.log` |
| negative cases | see below | 1 each (refusal expected) | `qual-logs/negative-cases.log` |

`make platform-hosted-topology-check` runs the agentd check, then parses all
nine manifests with PyYAML and again with the checker's own builtin parser, and
reports the new edge `ok` under both:

```
ok  Deployment layerx-testnet/layerx-gateway -> layerx-faucet-public.layerx-testnet.svc.cluster.local:443 [env LAYERX_GATEWAY_FAUCET_URL]:
    pod Deployment layerx-testnet/layerx-faucet:9443/TCP ingress admitted by layerx-faucet-public-ingress;
    egress admitted by layerx-gateway (DNS via layerx-gateway)
```

The manifest assertion gate loads both edited manifests with PyYAML and with the
builtin parser and checks, under each, that the projected volume names exactly
`{provisioning: token}`, that both new mounts are `readOnly`, that the three
variables hold the values the two services read, that both ends resolve to the
same Secret key, that the two policy rules exist, that bring-up both generates
and applies the Secret, and that neither manifest carries an inline token value.

Three negative cases confirm each piece is load-bearing, run in a throwaway copy
of the manifests under the scratch directory, which was deleted afterwards:

1. `layerx-faucet-public-ingress` reduced to the new `app=layerx-gateway` rule
   alone: the gateway edge stays `ok` while the testnet-control edge fails, so
   the new rule, not the `0.0.0.0/0` rule, admits the gateway.
2. the gateway egress rule to `app=layerx-faucet` removed:
   `FAIL ... egress NetworkPolicy layerx-gateway does not admit 9443/TCP to the callee`.
3. the named edge removed from `RESOLVED_EDGES`:
   `AssertionError: ('baseline resolved edges', "unnamed: ['Deployment layerx-testnet/layerx-gateway -> layerx-faucet-public...[env LAYERX_GATEWAY_FAUCET_URL]']", 'named but absent: []')`.

**Not run: the applied-cluster proof.** No Kubernetes cluster tooling is
installed on this host: `kind`, `kubectl`, `minikube` and `k3d` are all absent
(`qual-logs/cluster-capability.log`), and `platform/hosted/tests/beta-cluster.sh`
drives every step through `"$TOOLS_DIR/kubectl"` (line 142-144), so no pod was
ever scheduled with the new mounts and no live claim was made. The evidence above
is manifest level only, and `observation.3.7.9204` states that plainly rather
than implying a cluster run.

## Observations added

`spec/layerx-beta/qualification.kvx`, inside this lane's range:

| Id | Severity | Subject |
| --- | --- | --- |
| `3.6.9200` | note | The provisioning token mount that closes `6.5.4200`, and why one key is projected rather than the whole Secret. |
| `3.6.9201` | blocker | The `provisioning` role also admits `POST /v1/sessions`, so the gateway pod now holds an unused session-minting capability; the repair is a narrower identity role. |
| `3.6.9202` | note | The gateway-to-faucet wiring that closes `3.6.4301`: one shared Secret, three variables, two mounts, two policy rules. |
| `3.6.9203` | note | The named edge and the manifest, code and policy lines that authorise it. |
| `3.7.9204` | note | How the token is provisioned by bring-up, and exactly which proof did not run and why. |
| `3.6.9205` | note | `docs/wiki/HostedFaucet.md:297` now contradicts the manifest; outside this row's paths. |
| `3.6.9206` | note | Faucet network admission is keyed by peer address, so a proxied claim path shares one bucket per gateway pod. |

`spec/layerx-beta/report.md` was re-rendered by `tools/ci/beta-report.sh` in the
same commit and never edited by hand.


paxlabs-inc:
Rebased mechanically onto main 8b62a0f4f: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 3003a3bd8: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 417f84a8d (no file overlap with main since the merge base).

PR #258, opened 2026-09-11
Price asset pause and unpause in the version 2 fee schedule and bind their supply

paxlabs-inc:
## Issue

Asset pause (ordinal 2) and unpause (ordinal 3) are real transitions that native and hosted
admission accept, but the version 2 fee schedule could not price them and their receipts could
not carry a supply binding.

* `LXP_ASSET_FEE_PRICE_COUNT` was 8 and `asset_ordinals` in `src/protocol/lxp_fee.c` listed
  `1, 4, 5, 6, 7, 8, 10, 11`, so `lxp_fee_compute` answered `LXP_ERR_UNKNOWN_ACTIVITY` for
  ordinals 2 and 3 whenever `parameters->version == 2` — a node on that schedule refused to
  price an activity its own module executes (`[observation.3.7.2701]`, blocker).
* `lxp_receipt_validate_supply` allow-listed operations `1, 4, 5, 6, 7, 8, 10, 11` and answered
  `LXP_ERR_NON_CANONICAL` for everything else, so `lxp_ctx_bind_asset_supply` could not build a
  probe receipt for 2 or 3 and a decision was needed on whether the two belong in the list
  (`[observation.3.7.2702]`, note).
* `agent/crates/layerx-crypto` had `Payment::Pause` and `Payment::Unpause` but no `1-2.hex` or
  `1-3.hex` fixtures, so the crate's `VECTORS` table did not cover either ordinal.

## What changed and why

**The version 2 price table is widened to ten entries.** `asset_ordinals` reads
`1, 2, 3, 4, 5, 6, 7, 8, 10, 11` and `lxp_asset_fee_name` gains `fee.asset.pause` and
`fee.asset.unpause` in ordinal position. `lxp_fee_schedule` resolves both through the same
`lxp_gov_param_enact` call the other eight prices use, and `lxp_fee_compute` adds the resolved
price to `base_fee` alongside the per-encoded-byte, per-execution-unit and per-storage-unit
components metered for every other asset ordinal. Neither ordinal is given a constant, a zero or
a borrowed price, and an absent named parameter still fails the whole schedule with the
parameter-bounds refusal rather than pricing the ordinal at zero.

**The canonical encoding grows from 215 to 247 bytes**, with the price count at offset 86 reading
10 and the ten prices at `87 + 16i`. Both canonical lengths are named in `include/layerx/lxp_fee.h`
as `LXP_FEE_PARAMS_V1_BYTES` and `LXP_FEE_PARAMS_V2_BYTES`, and every site that repeated the
literals now derives from them: the encoder and decoder, the daemon fee-estimate response buffer,
the shared LXGB metadata generators for Rust and Python, the hosted gateway local harness and the
layerx-client fee-estimate reader, which also checks the price-count byte. A version 2 schedule
encoded at the old length no longer decodes, so any persisted governance `fee.schedule` value and
any genesis metadata artifact carrying one has to be rebuilt from the current generators.

**The supply-binding question is settled by reading what binding version 1 promises.** On an
asset-module receipt whose result code is `LXP_OK` and whose asset is non-zero, the receipt carries
`total_units_before` and `total_units_after`, and the validator recomputes the total the operation
implies: 10 is `before + amount`, 11 is `before - amount`, 1 is zero before and after, and
4, 5, 6, 7 and 8 are unchanged. Pause and unpause move no units, so the total they imply is the
unchanged one, and they join that group. With them absent the validator answered
`LXP_ERR_NON_CANONICAL` — the answer reserved for an operation the binding version cannot describe —
and made a binding impossible to construct for two operations whose supply effect is exactly known.
Admitting them adds a real equality check rather than removing one; reserved ordinal 9 stays
refused and no other arm of the switch was touched.

**Fixtures and tests.** `1-2.hex` and `1-3.hex` come from the crate's own encoder: the existing
`fixtures_roundtrip_and_bind_all_fields` asserts `payment.encode(ACTOR)? == payload` for every
vector, so both fixtures are proven to be what the encoder emits, and they also flow through the
disclosure-binding, monetary-field, invalid-shape and per-offset mutation tests. A new test pins
that pause, unpause and account_open share one 34-byte body while decoding to three distinct,
correctly typed payments. `tests/test_fees.c` prices all ten ordinals from a named parameter
table and pins the canonical layout of a version 2 schedule; `tests/protocol/lxp_test_receipts.c`
walks operations 2 through 8, requires `LXP_OK` on equal totals and `LXP_FATAL_SUPPLY_MISMATCH`
on any difference, and keeps the refusals for operation 9 and for a non-asset module.

Two documentation statements made false by the widening were corrected with it: the version 2
pricing sentence in `docs/wiki/Modules.md`, which said pause and unpause were payable only under
version 1 until the set was widened, and the priced asset ordinal list in the `lx_estimateFee`
description of the gateway OpenRPC document.

## Gates

Branch base `755e9b6cf`. The build-dependent gates ran at `4762fa84a`, which differs from the
head only by `spec/layerx-beta/qualification.kvx` and the report rendered from it; the ledger,
report and whitespace gates were re-run at the head.

| --- | --- | --- | --- |
| `make CC=gcc test-asset-registry` | 0 | `qual-logs/test-asset-registry.log` | `4762fa84a` |
| `make CC=gcc test` (aggregate) | 0 | `qual-logs/make-test.log` | `4762fa84a` |
| `CARGO_TARGET_DIR=/root/lx-shared-target/agent cargo test --manifest-path agent/Cargo.toml -p layerx-crypto` | 0 | `qual-logs/cargo-layerx-crypto.log` | `4762fa84a` |
| `make CC=gcc test-fees` | 0 | `qual-logs/test-fees.log` | `4762fa84a` |
| `make CC=gcc test-receipts` | 0 | `qual-logs/test-receipts.log` | `4762fa84a` |
| `make CC=gcc test-genesis` | 2 | `qual-logs/test-genesis.log` | `4762fa84a` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `373fa75de` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `373fa75de` |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | `373fa75de` |
| ledger block verification against the base | 0 | `qual-logs/kvx-verify.log` | `373fa75de` |

`make CC=gcc test-fees` and `make CC=gcc test-receipts` were added to the row's gate list because
the aggregate `test` target reaches neither, so without them nothing in the required set would
have executed the changed tests.

`make CC=gcc test-genesis` exits 2 for a cause this change does not touch: the genesis manifest
test requires asset ordinal 9 to be unresolvable while the asset module advertises
`LX_ASSET_WITHDRAW`. Neither file appears in this branch's changed-file list and both are byte
identical to the base. Recorded, not repaired.

The hosted gateway local harness and the `platform/hosted/core`, `authority` and `agent-boundary`
node harnesses consume `tests/support/lxgb_metadata.rs` and were updated for the new length, but
were not run here: they need a cluster or the native binaries whose genesis builder is already
recorded as refusing their LXGB request.

## Observations added

| Id | Severity | Subject |
| --- | --- | --- |
| `3.7.9400` | note | The widened price table, the 215 to 247 byte canonical encoding change and the artifacts that have to be rebuilt because of it. Closes `3.7.2701`. |
| `3.7.9401` | note | What supply binding version 1 promises, and the decision to admit operations 2 and 3 into the unchanged-supply group on that basis. Answers `3.7.2702`. |
| `3.7.9402` | suspect | Pause and unpause receipts still carry no binding: `module_execute` returns before the bind call for those two ordinals, in asset module code this change does not own. |
| `3.7.9403` | blocker | `make CC=gcc test-genesis` exits 2 on unmodified files — the genesis side of the withdraw dispatch gap. |
| `3.7.9404` | note | The aggregate `test` target reaches none of `test-fees`, `test-receipts`, `test-asset-registry` or `test-genesis`. |


paxlabs-inc:
Rebased mechanically onto main 755868836: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 159df9c76: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 373fa75de (no file overlap with main since the merge base).

PR #260, opened 2026-09-11
Tie every copy of the LNI version to the schema and refuse an old minor as unsupported

paxlabs-inc:
## Issue

The LNI version constant existed in four independent copies with nothing tying them to each other or to `agent/schema/lni/v1.kvx` (observation 6.1.8602): the daemon enum at `cmd/layerxd/lxp_daemon_lni.c:47-48`, the native client enums at `tests/daemon/lxp_test_program_admission.c:50-51` and `tests/test_daemon_lni_admission.c:47-48`, the literal the payment client pins at `tests/daemon/lxp_test_pay1.c:126`, and the `Version` constants and `SCHEMA` of `agent/crates/layerx-client`. That is how the 1.4 against 1.5 split arose, and it would recur at the next minor.

The same row settles observation 6.1.8603: `send_asset_read` and `send_fee_estimate` answered a below-5 minor with `LXP_ERR_NON_CANONICAL` where every other version gate in the same file uses `LXP_ERR_VERSION_UNSUPPORTED`.

## What changed and why

**One parity check, driven from the schema document.** `tools/ci/lni-version-parity.sh` reads `major` and `minor` from the `[schema]` section of `agent/schema/lni/v1.kvx` and requires the same pair from every copy. For the client crate it additionally requires the `SCHEMA` initialiser to build at the schema minor, the declared `Version::V1_*` constants to be exactly the revisions zero through that minor (no more, no fewer), and the `include_str!` argument to resolve to the same file the version was read from, so the crate cannot embed one document and declare another. Every extraction requires exactly one declaration it can read, so a declaration reformatted beyond the pattern's reach fails the check loudly instead of passing quietly. **No version value changed anywhere**; every copy already held 1.5.

The check is wired as `make test-lni-version-parity` and the aggregate `test:` target depends on it, the way the module suites were wired, so a split fails `make test` before anything is compiled.

**Typed refusal for the two 1.5 reads.** `send_asset_read` and `send_fee_estimate` folded their minor gate into the same condition as their payload validation, so a client that legitimately speaks 1.4 and asks for an asset read was told its bytes were malformed rather than that the message does not exist at its version. Both now select the result the way the four version gates beside them do: `request->minor < 5U` chooses `LXP_ERR_VERSION_UNSUPPORTED` and every other branch of the same condition keeps `LXP_ERR_NON_CANONICAL`, so a genuinely malformed payload at minor 5 is still refused as non-canonical. No existing assertion pinned either result, so nothing was relaxed.

**Tests against a real daemon.** `typed_read_version_gate` in the native admission client sends a canonical asset read and a canonical fee estimate at minor 4 and requires refusal class 1 with `LXP_ERR_VERSION_UNSUPPORTED`; sends each with a truncated payload at minor 5 and requires refusal class 1 with `LXP_ERR_NON_CANONICAL`; then sends both canonically at minor 5 and requires the typed responses (tags 33 and 35), each carrying format version 1. The positive half is the first real-daemon coverage these two tags have had. The below-version request follows a named `TYPED_READ_MINOR` rather than the client's own pin, so a future bump does not silently stop exercising the gate.

## Gates

| --- | --- | --- | --- |
| `tools/ci/lni-version-parity.sh` (nine deliberate divergences in scratch copies, unaltered copy passes) | 1 on each divergence, 0 unaltered | `qual-logs/lni-version-parity-negative.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc test-lni-version-parity` | 0 | `qual-logs/lni-version-parity.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc --dry-run test` reaches `tools/ci/lni-version-parity.sh` | 0 | `qual-logs/aggregate-test-wiring.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc build build/bin/layerxd build/bin/layerx-genesis-build` | 0 | `qual-logs/build.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc test-program-admission` (shipped genesis metadata fixture) | 2 | `qual-logs/test-program-admission-shipped-fixture.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc test-program-admission` (fixture corrected locally, uncommitted) | 0 | `qual-logs/test-program-admission.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc test-program-admission` with the daemon refusal codes reverted | 2 | `qual-logs/test-program-admission-before-refusal-change.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make CC=gcc test-program-simulate` | 0 | `qual-logs/test-program-simulate.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `gcc -Werror -Wall -Wextra -Wconversion ... -c tests/daemon/lxp_test_pay1.c` | 0 | `qual-logs/pay1-compile.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `make beta-ledger-check` | 0 | `qual-logs/beta-ledger-check.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `tools/ci/beta-report.sh --check` | 0 | `qual-logs/beta-report-check.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |
| `git diff --check` | 0 | `qual-logs/git-diff-check.log` | `579328d21a43293c7ab8f20cc2a2876aba0e54bf` |

`make CC=gcc test-program-admission` cannot reach a daemon on this base: `platform/hosted/node/bootstrap.sh` stops at `bootstrap: layerx-genesis-build refused the genesis request` with the shipped `tests/support/lxgb_metadata.py` emitter, which is the custody record shape recorded as observation 6.1.5801 and owned by another change. `layerx-genesis-build` links `cmd/layerx-genesis` and `build/liblayerx.a` only and contains none of the files changed here. The gate was therefore run twice: once with the shipped fixture (exit 2, no client run) and once with the corrected emitter applied locally and left uncommitted, exactly as observation 6.1.8600 describes. `tests/support/lxgb_metadata.py` is unmodified in this change. Both runs are in the table and the situation is recorded as observation 6.1.9102.

## Observations added

| Id | Severity | Subject |
| --- | --- | --- |
| `6.1.9100` | note | The parity check, what each copy must satisfy, the ten scratch-copy cases, and why the `Version::V1_4` peer pins elsewhere are deliberately out of scope. Settles 6.1.8602. |
| `6.1.9101` | note | The two corrected refusal codes, the assertions that now pin them and the positive responses, and the run that fails against the previous code. Settles 6.1.8603. |
| `6.1.9102` | blocker | `make CC=gcc test-program-admission` is red on this base for the shipped genesis metadata fixture, with both runs as evidence. |

## Files changed

- `tools/ci/lni-version-parity.sh` (new)
- `Makefile` (one parity target and its `test:` prerequisite)
- `cmd/layerxd/lxp_daemon_lni.c` (the two refusal codes)
- `tests/daemon/lxp_test_program_admission.c` (the refusal and response assertions)
- `spec/layerx-beta/qualification.kvx`, `spec/layerx-beta/report.md`


paxlabs-inc:
Rebased mechanically onto main dfd9618fe: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 601ac0c36: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 579328d21 (no file overlap with main since the merge base).

PR #265, opened 2026-09-11
Split the payment disclosure decoder into per-activity helpers

paxlabs-inc:
## Summary

Strict clippy over the agent workspace was red on main before it reached any crate under test: `agent/Cargo.toml` denies `clippy::pedantic` workspace-wide, which bounds a function at a hundred lines, and `Payment::decode` in `agent/crates/layerx-crypto/src/payments.rs` had grown to 116 source lines (lines 403-518 on main), so `layerx-crypto` failed to compile under clippy and every agent-workspace clippy invocation stopped there.

This change restores the bound by decomposition, not suppression. The four long match arms of `decode` now live in named helpers beside the existing `decode_grant`: `decode_registration` (payments.rs:487), `decode_receive` (:500), `decode_program_transfer` (:526) and `decode_program_account` (:544). `decode` itself is 55 source lines (403-457). No `#[allow]` was added, the workspace deny is untouched and the hundred-line bound is unchanged. Byte-level behaviour is identical: each helper keeps the field evaluation order of the arm it replaces, `decode_receive` keeps the short-circuiting frame check (the second `u16` is read only when the first equals `0x5201`), `decode_program_transfer` keeps the zero and 256 leg bounds ahead of any leg read, `decode_program_account` keeps the `LXPA1` marker check, and `decode` keeps the version prefix guard, the unsupported-activity refusal, `d.finish()`, `validate` and the re-encode comparison that closes the round trip. Every former in-arm `return bad()` is now an `Err` the `?` operator propagates with the same `DisclosureError::MalformedPayload` value.

Five tests pin the boundaries the helpers now own, each payload derived from an existing fixture or the encoder (no fixture was added): an asset body is refused for any version prefix other than one; the receive frame marker and its ten-field count are rejected when altered; the programs transfer leg count is refused at zero and above 256; the programs account seed marker is refused on any flipped byte; an unregistered activity names its own type in the refusal.

Three ledger observations under task 5.1 record the decomposition and the two files outside this change's paths that still fail the workspace-wide deny once `layerx-crypto` compiles cleanly (see Follow-ups). The report is rendered from the ledger.

## Files changed

- `agent/crates/layerx-crypto/src/payments.rs` — `Payment::decode` split into `decode_registration`, `decode_receive`, `decode_program_transfer`, `decode_program_account` (decode path only; 73 insertions, 65 deletions).
- `agent/crates/layerx-crypto/tests/payments.rs` — five new tests at lines 363-461 (102 insertions, 0 deletions).
- `spec/layerx-beta/qualification.kvx` — observations 5.1.9900 (note), 5.1.9901 (blocker), 5.1.9902 (blocker) appended after main's last block; main's content is byte-identical (24 lines added, 0 deleted).
- `spec/layerx-beta/report.md` — rendered by `tools/ci/beta-report.sh` from the ledger (observation counts and ledger digest).

## Design decisions

- Decompose along the existing per-activity match arms so each helper has exactly one caller in `decode` and the wire layout of every payload is untouched; `d.finish()`, `validate` and the re-encode comparison remain in `decode` as the single closing check.
- Helpers return `Result<Payment, DisclosureError>` (or `Result<Registration, _>`) and use `?`, so the former in-arm `return bad()` becomes a propagated `Err` with the identical error value; no new error variant, no fallback.
- The new tests reuse the fixture table and `hex` helper already in the test file and assert the exact bytes they mutate (`5201000a` at the head of `1-6.hex`, the `0002` leg count at offset 32 of `9-5.hex`, the `LXPA1` marker at offset 32 of `9-6.hex`, fixture lengths 258 and 77), so a fixture change is caught rather than silently retargeted.
- Every claim in the three ledger blocks was re-verified against the source, the fixtures and the retained logs before publication: function line numbers, the 116/55 line counts, the test line range, the 49 passing tests including the five new ones, the six lint sites in `tests/support/lxgb_metadata.rs`, the two in `agent/crates/layerx-client/tests/supply_receipt.rs`, the thirteen failing `layerx-agentd` targets, the nine workspace members covered by the keep-going run, and the exact command including `--locked`. No block was added or altered.
- The two files outside this change's paths are recorded as blockers 5.1.9901 and 5.1.9902 with the exact lint, line and behaviour-preserving repair, not patched here.

## Static checks

Each command was run from the worktree root against the pushed head `711434037`; exit codes as listed.

| Command | Exit |
| --- | --- |
| `rustfmt --edition 2021 --check agent/crates/layerx-crypto/src/payments.rs agent/crates/layerx-crypto/tests/payments.rs` | 0 |
| `git diff --check origin/main` | 0 |
| `git show origin/main:spec/layerx-beta/qualification.kvx > qual-logs/main.kvx && python3 kvx_verify.py qual-logs/main.kvx spec/layerx-beta/qualification.kvx` (block-by-block: `missing=[] differ=[]`, extra = the three 5.1.99xx blocks) | 0 |
| `git diff origin/main --numstat -- spec/layerx-beta/qualification.kvx` (`24 0`) | 0 |
| `bash tools/ci/beta-report.sh` (render; produced the committed `spec/layerx-beta/report.md` unchanged) | 0 |
| `bash tools/ci/beta-report.sh --check` | 0 |
| `bash tools/ci/beta-ledger-check.sh` (48 gate records, 1346 observation records) | 0 |
| `git merge-base HEAD origin/main` equals `origin/main` (`b3e57a437`; no rebase was needed) | 0 |

No build, compile, clippy or test command was run against this head. The three ledger blocks cite logs retained under the lane's untracked `qual-logs/` from an earlier run on this branch at revision `30c605d50` (same tree for both Rust files as the pushed head; merge base `dfd9618fe`, which already carried the merged genesis metadata fixture change) in a lane-private target directory: `cargo clippy --locked --manifest-path agent/Cargo.toml -p layerx-crypto --all-targets -- -D warnings` exit 0; `cargo test --locked --manifest-path agent/Cargo.toml -p layerx-crypto` exit 0, 49 tests passed including the five added here; `cargo clippy --locked --manifest-path agent/Cargo.toml --workspace --all-targets -- -D warnings` exit 101; the same command with `--keep-going` in an empty target exit 101, reporting clippy errors only in `tests/support/lxgb_metadata.rs` and `agent/crates/layerx-client/tests/supply_receipt.rs`. Between that run's base and the current base, main changed `agent/crates/layerx-sdk/src/rpc.rs` and `agent/crates/layerx-sdk/src/rpc_subscription.rs` (WebSocket unsubscribe and resume cursor), which the keep-going run did not see; both files named as blockers are byte-identical between the two bases and the pushed head.

none

## Follow-ups

- The row's success criterion "Agent workspace is clippy-clean under the declared lints" is not met by this change alone, and cannot be met from its owned paths: with `layerx-crypto` clean, the workspace gate fails on `tests/support/lxgb_metadata.rs` (clippy::format_collect at line 14, clippy::cast_possible_truncation at 18, 36 and 48, clippy::expect_used at 22 and 78; included by thirteen `layerx-agentd` integration targets through `agent/crates/layerx-agentd/tests/support/real_authority.rs`) and on `agent/crates/layerx-client/tests/supply_receipt.rs` (clippy::expect_used at lines 32 and 33). Each has a behaviour-preserving repair (a `write!` fold, `u16::try_from` with an explicit refusal, and the let-else form the receipt test already uses at lines 12 to 26); the exact requests are in blocks 5.1.9901 and 5.1.9902.
- Once those two files are repaired, `cargo clippy --locked --manifest-path agent/Cargo.toml --workspace --all-targets -- -D warnings` should be run through `tools/ci/beta-qualify.sh` against the merged revision so it lands in the ledger as a `[gate.5.1.n]` record; gate records are written only by that tool after an executed run, so none is added here.


paxlabs-inc:
Rebased mechanically onto main a3b941967: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head fad755cd9: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head 711434037 (no file overlap with main since the merge base).

PR #271, opened 2026-09-11
Retain the genesis request beside the genesis artifacts and speak LNI 1.5 in the availability harness

paxlabs-inc:
## Summary

Row G14. `make CC=gcc test-daemon-availability` could not pass against a real node for two recorded reasons (observations 6.1.8604 and 6.1.8601): `assert_genesis_request` read `data/work/genesis-request.lxgb`, which `platform/hosted/node/bootstrap.sh` removes with the whole scratch directory at the end of every successful bootstrap, and the harness pinned `Version::V1_4` at seven sites while the crate's `SCHEMA` and the daemon declare 1.5.

The bootstrap now retains the LXGB v2 request it built genesis from as `genesis/genesis-request.lxgb`, beside the manifest, snapshot, registration request and descriptor derived from it, before it removes the scratch directory that also held the genesis signer key. The harness reads that retained copy, requires the scratch directory to be absent from the completed data directory, and speaks `Version::V1_5` everywhere it previously spoke 1.4. Every existing length, envelope-version, parameter-count, guarantor-order, key-prefix, bond and guarantor-key assertion is unchanged; nothing was deleted or made conditional.

## Files changed

- `platform/hosted/node/bootstrap.sh` (genesis-request retention hunk only): after the four genesis artefacts are confirmed present, `mv "$REQUEST" "$GENESIS_DIR/genesis-request.lxgb"` (lines 504-505); the header's "Outputs under DATA_DIR" list names the retained request and the `usage()` slice grows by the one header line it prints (`2,88p` to `2,89p`). `rm -rf "$DATA_DIR/work"` at the end of the bootstrap is unchanged.
- `agent/crates/layerx-client/tests/availability_daemon.rs`: the seven `Version::V1_4` pins at lines 21, 47, 95, 117, 177, 194 and 202 read `Version::V1_5`; `assert_genesis_request` asserts `data/work` no longer exists, then reads `data/genesis/genesis-request.lxgb`; its doc comment says where the request is read from and why.
- `docs/wiki/HostedNode.md`: one row for `genesis/genesis-request.lxgb` in the bootstrap outputs table.
- `spec/layerx-beta/qualification.kvx`: observations `6.1.9000` (note: the retention decision, the pin moves, the wire-level reasoning, and the plain statement that build and test were not run here) and `6.1.9001` (suspect: the 6.1.8605 poll decision) appended after main's blocks, which are kept byte for byte.
- `spec/layerx-beta/report.md`: re-rendered from the ledger by `tools/ci/beta-report.sh`, not hand-edited (observation counts only).

## Design decisions

- Retain rather than relocate the read. The bootstrap decides what a node keeps and the harness reads what it leaves; the request carries public guarantor keys, the asset id, the metering and fee parameters and the metadata suffix and no secret, so keeping it at the mode the bootstrap already gave it (0600, under the 0700 data directory) widens nothing. The scratch directory is still removed because it held the genesis signer key; the request is moved out of it only after `layerx-genesis-build` has succeeded and its artefacts are confirmed. The supervisor reset discards the retained copy with everything else through `--force`.
- The `mv` sits after the artefact-presence loop, so a refused genesis never leaves a half-populated genesis directory holding a request no artefact was built from.
- The extra assertion that `data/work` is gone tests the second half of the same decision (what the bootstrap must not leave) and is the only assertion added.
- Wire-level effect of the pin moves, checked against the daemon: `decode_envelope` (`cmd/layerxd/lxp_daemon_lni.c:1179-1181`) accepts any minor up to `LNI_VERSION_MINOR = 5`; `perform` still sends the node-info request at minor 0 (`handshake.rs:132-135`); every daemon response carries the daemon revision and each client path the harness uses (`batch::lookup`, `evidence::exchange`, `availability::fetch*`) compares only the major. The single minor-5 selector branch, `wait_publication` in `send_receipt` (`lxp_daemon_lni.c:2148-2152`), is on the receipt-lookup message this harness never sends; `send_batch_header` and `send_availability` have no minor branch, so the 10-byte batch-header selector for batch 1 (which ends in `0x01`) and the 9/17/33-byte availability selectors are read identically at minor 4 and 5.
- `platform/hosted/node/tests/bootstrap-test.py:38-45` opens the request while the bootstrap is still running and keeps the descriptor, so the move does not change what it reads.
- Observation 6.1.8605 (200 x 10 ms receipt poll where siblings pause 50 ms) is left exactly as written. The row asked to align it only if the shorter bound is shown not to be what the contract requires; `agent/schema/lni/v1.kvx:50` says minor 5 permits a trailing `wait_publication` byte under which the node waits on publication events until the request deadline and "never a fixed polling interval", so neither 2 s nor 10 s is a contract bound and the right change is the wait flag, not a different pause. That file is outside this row's owned paths; the request is recorded in observation 6.1.9001 and in the lane's untracked NEEDS.md.
- No asset_read or fee_estimate probes were added: the row asks for the pins to move so the 1.5 request set becomes reachable, and payload shapes that were never exercised against a real daemon cannot be checked without running one.

## Static checks

| Command | Exit |
| --- | --- |
| `bash -n platform/hosted/node/bootstrap.sh` | 0 |
| `rustfmt --edition 2021 --check agent/crates/layerx-client/tests/availability_daemon.rs` | 0 |
| `git diff --check` | 0 |
| `bash tools/ci/beta-ledger-check.sh` | 0 (48 gate records, 1345 observation records) |
| `bash tools/ci/beta-report.sh` then `bash tools/ci/beta-report.sh --check` | 0 and 0 |

No build, compile or test was run in this lane and no gate result is claimed. Gates for the owner's build: `make CC=gcc test-daemon-availability` three consecutive times (all must exit 0; observation 6.1.8605 records the run-to-run timing sensitivity of the native client's receipt poll, unchanged here), `make beta-ledger-check`, `make test-lni-version-parity`, and `make platform-test-node` for the bootstrap change.

none

## Follow-ups

- `tests/daemon/lxp_test_program_admission.c` (outside owned paths): send the receipt lookup as a 34-byte selector with `wait_publication` set and let the request deadline bound the wait, in place of the 200 x 10 ms loop (observation 6.1.9001).
- With the harness at 1.5, `layerx_client::payments::{list_assets, get_asset, estimate_fee}` can be exercised against the real daemon in a further probe stage.
- `platform/hosted/node/tests/bootstrap-test.py` still expects an LXGB envelope version of 1 and a request without the metadata suffix the bootstrap now requires (pre-existing, adjacent to the ledger's earlier `--genesis-metadata` record); the other line citations in the HostedNode.md outputs table are also pre-existing stale.


paxlabs-inc:
Rebased `lane/day-g14` onto `origin/main` (`bb5f5481f`); new head `d58add8df`. The branch keeps its single commit; nothing was amended or squashed beyond the rebase itself.

Conflicts and how they were resolved:

- `docs/wiki/HostedNode.md` — auto-merged cleanly by git, then re-read. Main's changes to the cluster-shape tables (treasury-signer, guarantor, agentd-boundary and Human containers; `layerx-agentd` and `layerx-human` Services; updated deployment.yaml line references) are kept verbatim, and this branch's single row for `genesis/genesis-request.lxgb` in the "Bootstrap artefacts" table is retained after `genesis/genesis.registration`. No conflict markers remain.
- `spec/layerx-beta/qualification.kvx` — auto-merged. Verified with `kvx_verify.py` against `origin/main`'s file: `missing=[] differ=[] extra=['observation.6.1.9000', 'observation.6.1.9001']`; `git diff origin/main --numstat` shows `15 0` (zero deletions), so main's last block keeps its `severity` line and only this branch's two existing blocks follow it. No new blocks were added.
- `spec/layerx-beta/report.md` — the only textual conflict. Resolved by taking `origin/main`'s version and re-rendering it with `tools/ci/beta-report.sh` (exit 0); `tools/ci/beta-report.sh --check` exits 0 against the result.

Static checks run (no build or test in this lane):
- `rustfmt --edition 2021 --check agent/crates/layerx-client/tests/availability_daemon.rs` — exit 0
- `bash -n platform/hosted/node/bootstrap.sh` — exit 0
- `git diff --check origin/main HEAD` — exit 0
- `tools/ci/beta-report.sh` / `tools/ci/beta-report.sh --check` — exit 0 / 0

PR #273, opened 2026-09-12
Start a real layerxd from the hosted node harnesses

paxlabs-inc:
## Summary

The hosted node test harness was red on main for the three reasons recorded at observations 6.1.4802, 6.1.4803 and 6.1.4804. `bootstrap.sh` requires `--genesis-metadata` and neither `node-test.sh`, `bootstrap-test.py` nor the sequencer arguments in `deployment.yaml` passed it; the probe crate's `Cargo.lock` was stale so `cargo build --locked` refused it; and `platform/hosted/registry/tests/native_deployment.rs` wrote a version 1 genesis request without Asset records, which the daemon refuses with `LXP_ERR_ASSET_MISMATCH` (`-402`) when `collect_assets` finds no committed asset (`cmd/layerxd/lxp_daemon_process.c:595`). Each cause is fixed in the code or configuration. No assertion was edited.

Following the row addenda: the `bootstrap.sh:450` threshold bound was already fixed by PR 239 and is not touched; the `bootstrap.sh:202` single-guarantor metadata bound now follows the certificate threshold; `LAYERX_TEST_SOCAT_BIN` has a real default in the hosted node harness.

## Files changed

- `platform/hosted/node/tests/node-test.sh`: resolves socat from `LAYERX_TEST_SOCAT_BIN` (refusing `socat_invalid` when it is not an executable file) or from `PATH` (refusing `socat_missing`), and passes `--socat` to both supervisors; derives the treasury public key from the generated seed with the same PKCS#8 wrapper `bootstrap.sh` uses; builds the LXGB v2 metadata suffix with `tests/support/lxgb_metadata.py` (running its pinned-vector `check()` first) over the beta asset, kept under `$WORK` outside the data directory so the supervisor reset can bootstrap again from it; passes `--asset` and `--genesis-metadata` to the sequencer supervisor; checks `node.env` asset id and treasury public key; requires `jq` and `python3`, which it already used or now uses.
- `platform/hosted/node/tests/bootstrap-test.py`: loads `tests/support/lxgb_metadata.py` through `importlib.util` (the way `node-test.sh` does, so no `sys.path` mutation precedes an import), builds the same metadata suffix per run from a treasury public key derived with openssl and passes `--asset` and `--genesis-metadata`. Its two assertions at lines 84-85 are left as written (see Follow-ups).
- `platform/hosted/node/deployment.yaml`: sequencer arguments gain `--genesis-metadata /run/layerx/genesis/metadata.lxgb`; that path is a `subPath` mount of key `metadata.lxgb` of ConfigMap `layerx-node-genesis-metadata` (one volumeMount, one volume).
- `platform/hosted/node/tests/probe/Cargo.lock`: relocked for the workspace members only (14 packages added, unused `base64` removed).
- `platform/hosted/registry/tests/native_deployment.rs`: includes `tests/support/lxgb_metadata.rs` the way `platform/hosted/core/tests/boundary.rs` and the authority and agent-boundary harnesses do; `genesis_request` writes version 2 and appends the shared suffix issued to the treasury key; `build_genesis` takes the treasury public key; the single caller in `start_cluster` passes it.
- `platform/hosted/node/bootstrap.sh`: the metadata length bound moves from argument validation to directly after `GUARANTOR_COUNT` is read from the settlement document (still before any directory is created) and becomes `16384 - 314 - 81 * GUARANTOR_COUNT`, which equals the previous `15989` for one guarantor; the refusal names the observed length, the admitted range and the guarantor count.
- `docs/wiki/HostedNode.md`: metadata bound, the manifest mount, and the node-test behaviour; every `node-test.sh` and `bootstrap.sh` line reference in the Tests and genesis request sections is re-derived from the changed files.
- `spec/layerx-beta/qualification.kvx`: observations 6.1.7700 (bootstrap-test version 1 assertions), 6.1.7701 (boundary.rs still lacks a socat default), 6.1.7702 (bring-up must publish the genesis metadata ConfigMap). Each states that build and test were not run in this lane.

## Design decisions

- The metadata file in the manifest is mounted with `subPath` because a plain ConfigMap volume presents `metadata.lxgb` as a symlink into `..data`, which `bootstrap.sh:211` refuses (`! -L`). The ConfigMap is not `optional`: the shipped configuration must not bootstrap without the authoritative Asset records, so a missing map fails at pod start with a visible mount error instead of silently. Row scope was "daemon args only"; the volume and volumeMount are the minimum that makes the argument name an existing file, and are called out here.
- The metadata bytes are authoritative (real issuer, salt, cap) and are not invented in the manifest or the image; the harnesses generate throwaway ones for throwaway keys, exactly as `signer-test.py`, `tests/daemon/pay1.py` and the core boundary harness already do.
- The probe lock was repaired with `cargo update --workspace --offline`, which performs the same workspace-only re-resolution `cargo build` would (row said never `generate-lockfile`; the owner builds in this workflow, so `cargo build` itself was not run).
- The issuer key for the registry harness metadata is the cluster treasury key, matching the supervised cluster in `boundary.rs:2302-2309`. The daemon does not require records to match the request asset, but both use the same random `asset` so the treasury `asset` reads resolve.
- `bootstrap-test.py` derives the treasury public key by piping the PKCS#8-wrapped seed to `openssl pkey` over stdin; the keys are random per test run and discarded.

## Static checks

- `bash -n platform/hosted/node/tests/node-test.sh` — exit 0
- `bash -n platform/hosted/node/bootstrap.sh` — exit 0
- `python3 -m py_compile platform/hosted/node/tests/bootstrap-test.py` — exit 0
- `python3 -c "import yaml; list(yaml.safe_load_all(open('platform/hosted/node/deployment.yaml')))"` — exit 0
- `rustfmt --edition 2021 --check platform/hosted/registry/tests/native_deployment.rs` — exit 1; the only diff reported is inside the included, unchanged `tests/support/lxgb_metadata.rs` (its `public_hex` chain), which is pre-existing and shared with the core, authority and agent-boundary harnesses; the hunks of `native_deployment.rs` itself produce no diff
- `cargo update --workspace --offline --manifest-path platform/hosted/node/tests/probe/Cargo.toml` — exit 0 (lockfile only)
- `git diff --check origin/main HEAD` — exit 0
- ledger: every block of `origin/main` is present and identical (`kvx_verify` missing=[] differ=[]); the three new blocks carry exactly `task, file, symbol, observed, assumption, severity` with quoted values
- `make beta-ledger-check` and `tools/ci/beta-report.sh --check` were not run (no `make` in this lane)

`make CC=gcc platform-test-node`, `cargo test -p layerx-platform-registry --test native_deployment -- --test-threads 1`, `make beta-ledger-check` and the native binaries were not built or run here; the owner runs them once the work is on main.

- `spec/layerx-beta/report.md`: `tools/ci/beta-report.sh` (then `tools/ci/beta-report.sh --check`), because the ledger gained three observation blocks
- ConfigMap `layerx-node-genesis-metadata` (key `metadata.lxgb`) must be published by cluster bring-up from the authoritative LXGB v2 metadata before `platform/hosted/node/deployment.yaml` is applied; not part of this repository's tracked files

## Follow-ups

- `platform/hosted/node/tests/bootstrap-test.py:84-85` still assert `len(request) == 314 + 81 * count` and `request[:5] == b'LXGB\x01'`; `bootstrap.sh:460-484` writes version byte 2 plus the metadata suffix since 72c986271, so with the metadata now supplied `test_real_genesis_sets` reaches those assertions and fails there. Left unedited per the row ("never edit assertions"); recorded as observation 6.1.7700. Not part of `make platform-test-node`.
- `platform/hosted/core/tests/boundary.rs:2253-2256` still panics without `LAYERX_TEST_SOCAT_BIN` (outside owned paths); observation 6.1.7701.
- `platform/hosted/tests/beta-cluster.sh` must publish `layerx-node-genesis-metadata` (outside owned paths); observation 6.1.7702.
- `tests/support/lxgb_metadata.rs` is not rustfmt-clean (pre-existing, shared).


paxlabs-inc:
Rebased mechanically onto main 98463fcf6: the only conflict was the appended ledger blocks, resolved as main's file followed by this branch's blocks (block-for-block equivalence with both sides verified), and spec/layerx-beta/report.md was re-rendered with tools/ci/beta-report.sh against the merged ledger (tools/ci/beta-report.sh --check exit 0). At the new head 14e279d55: make beta-ledger-check exit 0, git diff --check clean. The code hunks are unchanged from the previously qualified head dc546457c (no file overlap with main since the merge base).

PR #274, opened 2026-09-12
Dispatch Asset withdraw through the real transition and bind pause and unpause supply

paxlabs-inc:
## Summary

Asset ordinal 9 (withdraw) was advertised in the module's `activity_types` but `module_execute` resolved the supply-binding asset id through a switch whose default arm returned `LXP_ERR_UNKNOWN_ACTIVITY`, so the real withdraw transition was unreachable through `lxp_kernel_dispatch` (observation 3.7.2700). This change routes ordinal 9 straight into `module_execute_impl` / `execute_withdrawal`, which already carries the journal rollback on a failed transfer set, the `LXP_REASON_WITHDRAWAL` debit into `system:paxeer-withdrawals` and the `withdrawal:` record. It also fixes the two neighbouring gaps found while reading the dispatch path: pause and unpause (2/3) receipts now carry the supply binding (M13 addendum b, observation 3.7.9402), and the kernel's ledger admission facts now record actor, verified key and the principal account snapshot for a withdraw so `lxp_kernel_withdraw_execution_sequence` can accept a dispatched withdraw instead of refusing every one with `LXP_ERR_CONTEXT_MISMATCH`. `tests/test_genesis_manifest.c` no longer asserts that the advertised ordinal 9 is unresolvable (addendum a, observation 3.7.9403). `tests/modules/test_asset_withdraw_activity.c` and its assertions are unchanged.

## Files changed

- `src/modules/asset/lx_asset_registry.c`: `module_execute` dispatches ordinal 9 to `module_execute_impl`; ordinals 2 and 3 take their asset id from the typed pause payload and fall through to `lxp_ctx_bind_asset_supply` like every other ordinal.
- `src/protocol/lxp_kernel.c`: `lxp_kernel_bind_ledger_admission` records actor, verified key, `account_present` and `next_sequence` for `LX_ASSET_WITHDRAW` as it already does for `LX_ASSET_ACCOUNT_OPEN`; `lxp_kernel_withdraw_execution_sequence` keeps every fact check and takes the signed `account_sequence` (enforced by `lxp_apply`) only when the admission was bound with no principal at all.
- `tests/test_genesis_manifest.c`: the state-commitment genesis check requires all eleven declared asset ordinals to resolve and undeclared ordinal 12 to refuse.
- `tests/modules/test_asset_registry.c`: new `test_supply_binding_dispatch` (register, pause, refused pause, unpause receipts through `lxp_kernel_dispatch`) and `test_withdraw_dispatch` (principal mismatch refused with state root unchanged; then debit 25 of 100, receipt without supply binding, committed `withdrawal:` record matching `lx_withdrawal_state_encode`; then `LXP_ERR_WITHDRAWAL_ALREADY_SETTLED`). `pause_submit` is a thin wrapper over `pause_submit_receipt`, which additionally hands back `ctx.ledger_receipt`.
- `docs/wiki/Modules.md`: asset paragraph declares withdraw (9), describes its payload, debit, record, admission facts and the absence of a supply binding.
- `spec/layerx-beta/qualification.kvx`: appended `[observation.3.7.9500]`.

## Design decisions

- Ordinal 9 bypasses the supply-binding step because `lxp_receipt_validate_supply` (src/protocol/lxp_receipt.c, outside this row) does not describe operation 9; a withdraw receipt therefore has `supply_binding_version == 0`, which the new test asserts. Extending the allow list is a follow-up rather than a silent change here.
- The withdraw admission gap is a kernel fact, not an asset-module fact, so `src/protocol/lxp_kernel.c` was edited outside the owned paths under the row's instruction that an incomplete withdraw transition is to be completed, never gated off. The change is confined to the `LX_ASSET_WITHDRAW` branch; other asset ordinals bind exactly as before.
- The no-principal fallback in the sequence resolver is taken only when `account_id`, `principal` and the account snapshot are all absent, which is what `lxp_kernel_dispatch` produces for a caller that resolved no principal; the daemon always resolves `agent:<did>:main` as principal (`cmd/layerxd/lxp_daemon_lni.c`), so a live withdraw takes the fully fact-checked path. A bound principal that is not the debited account still refuses.
- `tests/test_genesis_manifest.c` (outside owned paths) contradicted the row's success criterion by asserting the advertised ordinal 9 does not resolve; by inspection it is red on origin/main because `activity_types` advertises `LX_ASSET_WITHDRAW` and `lxp_kernel_module_for_activity` resolves it. The assertion was moved to undeclared ordinal 12, not deleted.
- Prior uncommitted work found in the worktree was inspected, saved to the scratchpad and discarded; the branch was rebuilt on origin/main.

## Static checks

- `git diff --check` — exit 0
- `python3 kvx_verify.py <origin/main qualification.kvx> spec/layerx-beta/qualification.kvx` — exit 0 (`missing=[] differ=[]`, one extra block `observation.3.7.9500`)
- `git diff origin/main HEAD --numstat -- spec/layerx-beta/qualification.kvx` — 8 added, 0 deleted

## Build and test

Not run in this lane. `make CC=gcc test-asset-withdraw-activity`, `test-asset-registry`, `test-genesis`, `test-fees`, `test-receipts`, `test`, `beta-ledger-check` are for the owner's build on main.

- None regenerated here. If the owner's `make CC=gcc test-native-withdrawal-proof` reports a vector mismatch, the checked-in vector is `human/crates/layerx-paxeer-client/tests/vectors/native-withdrawal-proof.json`; this change does not alter the proof encoding, so no mismatch is expected.
- `tools/ci/beta-report.sh --check` after the owner records gates.

## Follow-ups

- Fold `test-fees`, `test-receipts`, `test-asset-registry` and `test-genesis` into the aggregate `make test` (observation 3.7.9404).
- Admit operation 9 in `lxp_receipt_validate_supply` and `lxp_ctx_bind_ledger_receipt` if withdraw receipts should carry supply and ledger bindings.
- The hosted and native submission registries still refuse ordinal 9 with a reserved-operation refusal; `spec/layerx-protocol/spec.kvx` do_1 still calls ordinal 9 reserved; the version 2 fee schedule prices ten ordinals.


paxlabs-inc:
## Rebased onto main

`lane/day-m13` was replayed from `a963a26f` onto `cf26456d` (current `origin/main`). New head: `d02e9a30`.

A rebase had been left in progress against `59326b18`, which is eight commits behind current main, and its staged resolution of `spec/layerx-beta/qualification.kvx` added 6 lines where the commit adds 8: the textual merge had folded the closing `severity = "note"` line of main's final block into this branch's block and dropped the blank separator, so main's record lost a line. That attempt was discarded and the rebase redone against current main.

### Conflicts and how they were resolved

**`tests/test_genesis_manifest.c`** (content conflict, both sides rewrote the same block of `check_version`). Resolved by taking main's version whole, so the file is now byte-identical to `origin/main`. Main's version (PR 272, `f7bf0052`) is a strict superset of this branch's edit:

- it resolves every advertised Asset activity type by name, `LX_ASSET_REGISTER` through `LX_ASSET_BURN`, which includes `LX_ASSET_WITHDRAW` (ordinal 9) — the assertion this row's dispatch change exists to satisfy, and the one the pre-rebase file asserted as `ordinals[] = {1U .. 11U}`;
- it additionally requires `entry->module_id == LXP_MODULE_ASSET`, `entry->abi_version == lx_asset_module_iface()->abi_version` and `entry->activity_type_count == 11`;
- it refuses `(uint32_t)LXP_MODULE_ASSET << 16U` (ordinal 0) and `LX_ASSET_BURN + 1U`, which is `0x0001000c`, the same undeclared ordinal 12 this branch asserted as refused.

Nothing this branch asserted is dropped, and main's `#include "layerx/lx_asset.h"` at line 4 covers the named constants and `lx_asset_module_iface()`.

**`spec/layerx-beta/qualification.kvx`** (auto-merged without a marker, and wrong). The three-way merge appends this branch's `[observation.3.7.9500]` directly after main's final block with no blank line between them — 7 added lines where the commit adds 8 — which is the only block boundary in the file without a separator. The file was rebuilt as main's exact bytes followed by this branch's single existing block, separator restored. No block was added, renumbered or edited.

- `kvx_verify main.kvx spec/layerx-beta/qualification.kvx` → `blocks expected=1539 actual=1540 missing=[] differ=[] extra=['observation.3.7.9500']`
- `git diff origin/main --numstat -- spec/layerx-beta/qualification.kvx` → `8	0` (zero deletions)

**No conflict** in `docs/wiki/Modules.md`, `src/modules/asset/lx_asset_registry.c`, `src/protocol/lxp_kernel.c`, `tests/modules/test_asset_registry.c`. Main has not touched `src/` or `include/` since this branch's base, and each of those four blobs carries over identical to `a963a26f` (verified by comparing the index blob hashes against the original commit). The declarations the change depends on are unchanged on main: `lxp_activity_type_ordinal` in `include/layerx/lxp_activity.h`, `lxp_ct_is_zero` in `include/layerx/lxp_crypto.h`, and the `pause` arm of the typed payload used at `src/modules/asset/lx_asset_execution.h:236`.

**`spec/layerx-beta/report.md`** was taken from main and re-rendered, because the added observation changes the ledger digest the report records. `tools/ci/beta-report.sh` regenerated it in its own commit; only the `ledger_digest` row and the observation counts move (notes 717 → 718, total 1491 → 1492). The decision, the surface rungs and the stop conditions are unchanged.

### Note for the reviewer

Observation `3.7.9500` is carried over verbatim, as the ledger rule for a rebase requires. Its text describes the genesis-manifest assertion change, which main's PR 272 now carries, so after this rebase the branch no longer modifies `tests/test_genesis_manifest.c` while the block still names it. The statement remains true of the tree's final state; the block was left unedited rather than reworded.

### Static checks

| Command | Exit |
| --- | --- |
| `git diff --check origin/main HEAD` | 0 |
| `tools/ci/beta-report.sh` (render) | 0 |
| `tools/ci/beta-report.sh --check` | 0 |
| `kvx_verify` against `origin/main:spec/layerx-beta/qualification.kvx` | `missing=[] differ=[]` |

No Rust or shell sources changed, so `rustfmt --check` and `bash -n` have no targets. Build and test were not run in this lane.

PR #279, opened 2026-09-12
Execute enabled modules and recover authenticated handovers

paxlabs-inc:
Enabled Escrow, Budget, Stream, Service and Perps now execute through the daemon and guarantor. Epoch transitions publish deterministic maintenance receipts and events, preserve rollback safety, and recover durable state across restart. Native handover verifies the full predecessor history, replacement-key activation and checkpoint publication.

Validation at a6832bf81 passed the full native suite, guarantor build and TLS tests, strict Human Paxeer client Clippy, the historical publication regression, and the original two-peer handover and restart test with all 15 batches. The final 1e44b495a commit only restores the exact main ledger prefix and regenerates the report; ledger, report and whitespace checks pass.

The aggregate settlement journey still depends on the separate native withdrawal integration. Public handover trust consumers and final release qualification remain separate work.


paxlabs-inc:
The complete implementation ancestry is now on main through #305 (1ff52f9edbc6497cdc69473e2a93f6d0a1c2d25a). The only remaining commit on this branch rearranges qualification history and the generated report. Every observation is retained on main, including the already-applied removal of absolute checkout prefixes; merging this obsolete ordering would undo that reconciliation. Closed as superseded by the integrated implementation.

